PR #6, opened 2021-05-21
There were a few remaining glsl vec4

couleurs:
Converting to float4 for hlsl


patriciogonzalezvivo:
Amigo!!! What a nice surprise!

PR #7, opened 2021-05-24
mod doesn't exist in hlsl, need to use fmod

couleurs:


patriciogonzalezvivo:
I love you pibe. Keep it comming!

SabinT:
Not sure if this matters for the attached commit, but hlsl `fmod` produces a different output than glsl `mod` for negative numbers. Does it make sense for color values to be negative? If not, then non issue.

PR #9, opened 2021-06-03
fix define statements in sine.glsl

couleurs:


patriciogonzalezvivo:
thanks for all that <3 my friend

PR #10, opened 2021-08-07
Add HLSL version of grad4, initializer fix in random.hlsl

SabinT:
`grad4.glsl` didn't have a parallel HLSL version, adding that.
Note: 
I had to make the following change, because HLSL doesn't have a `lessThan` function:
```
// 1 - step(y, x) is the same as (x < y)

// GLSL: s = float4(lessThan(p, float4(0.0)));
s = float4(1 - step(float4(0, 0, 0, 0), p));
```

Also making some changes to `random.hlsl`, which had some `float{n}` constructors with a single argument.


patriciogonzalezvivo:
Thank you!

PR #12, opened 2021-08-17
Fix typo in hueShift.glsl/hlsl

SabinT:
Small typo in `hueShift.glsl` and `hueShift.hlsl`
There's an extra``` at the end :)


patriciogonzalezvivo:
Thank you!

PR #13, opened 2021-09-11
Add float version of HLSL mod

SabinT:
Completing the set :)

### DIFF
--- a/math/mod.hlsl
+++ b/math/mod.hlsl
@@ -12,7 +12,7 @@ license: |
 
 #ifndef FNC_MOD
 #define FNC_MOD
-float2 mod(float x, float y) {
+float mod(float x, float y) {
     return x - y * floor(x / y);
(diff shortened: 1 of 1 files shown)


math/mod.hlsl
```diff
@@ -11,6 +12,10 @@ license: |
 
 #ifndef FNC_MOD
 #define FNC_MOD
+float2 mod(float x, float y) {
```
patriciogonzalezvivo, line 15: I think there is a typo here. This one should return `float` and not `float2`. Right? 
Thank you so much for contributing!

SabinT: Oh whoops, thanks 😅 will fix.

SabinT: Done

patriciogonzalezvivo:
Thank you! This is great!

PR #21, opened 2022-03-01
fix small typo

arzafran:
esbuid -> esbuild

**Why?**
i was looking for esbuild support using CMD + F this typo kept from finding it.


patriciogonzalezvivo:
thanks!

PR #27, opened 2022-10-26
Convert raymarching GLSL to HLSL 

thnewlands:
Working in Unity 2021.3.8f1


patriciogonzalezvivo:
WOW! fantastic job! Thank you!

thnewlands:
Thank you! I just saw the rest of your changes. I had some distortion in unity at sheer angles to sphere using the direction from the camera to the vertex in world space but it felt like it I might have been missing something.

PR #33, opened 2022-11-12
add static TD version references

vectorsize:
This adds a new reference for TD, author details and small clarification


patriciogonzalezvivo:
Thank you!

PR #36, opened 2022-11-24
Update pbr.glsl

gzozulin:
Cannot be otherwise used as an import outside of Viewer; see usage on line 49


patriciogonzalezvivo:
Thank you

PR #37, opened 2022-11-24
Default SURFACE_POSITION to vec3(0.0)

gzozulin:
Same as before: if the PBR shader is included outside Viewer, this variable will be undefined. My understanding is that this mechanism allows to include in 'stock' configuration and override for custom use


patriciogonzalezvivo:
Thank you!

PR #38, opened 2022-12-22
feat: enhance sampleViewPosition

scarletsky:
In this PR, I support pass `depth` into `sampleViewPosition`.

On iOS 14-, the `WEBGL_depth_texture` extension has poor extension and we need to pack float to rgba8, and unpack it when we want to use it.

So it will be useful if we support passing `depth` directly.

One more thing, the old `sampleViewPosition` return type is wrong, and it has been fixed in this PR.


patriciogonzalezvivo:
Smart consideration! Thank you!

PR #39, opened 2022-12-25
Update boxBlur.glsl

gzozulin:
Parameter is missing


patriciogonzalezvivo:
Good catch! Thank you!

PR #43, opened 2023-02-10
fix: linearizeDepth wrong

scarletsky:
From https://learnopengl.com/Advanced-OpenGL/Depth-testing, the linearize depth formula is

```glsl
float ndc = depth * 2.0 - 1.0; 
float linearDepth = (2.0 * near * far) / (far + near - ndc * (far - near));
```

The code is missing `* far` part.


patriciogonzalezvivo:
Good catch! thank you

Issue #45, opened 2023-02-26
Triangular tiles - triTile


mathiasbredholt:
Just a quick suggestion here, I made a function for creating triangular tiles as I thought it could be a nice addition to the existing hex and square tiles. The code is based on FabriceNeyret2's [hexagonal tiling tuto](https://www.shadertoy.com/view/4dKXR3).

As the code probably can be improved I just included it here instead of filing a PR. If you prefer a PR @patriciogonzalezvivo, I can definitely do that as well. 

I'm not sure whether the barycentric coords should be optional as an bool argument or that we should just prefer one over the other. Alternatively, we could have `triTile()` and `triTileBaryCentric()`?

```
vec4 triTile(vec2 st) {
  // based on 
  // FabriceNeyret2 hexagonal tiling tuto
  // https://www.shadertoy.com/view/4dKXR3
  st *= mat2(1,-1./1.73, 0,2./1.73);         // conversion to hexagonal coordinates 
  vec3 g = vec3(st,1.-st.x-st.y);
  vec3 id = floor(g);                        // cell id
  g = fract(g);                              // diamond coords
  return length(g) > 1. ? vec4(g.xy, id.xy) : vec4(g.xy, id.yz);
  // barycentric coords, should this be optional?
  // return length(g) > 1. ? vec4(g.xy, id.xy) : vec4(1. - g.xy, id.yz);
}
```
Here's an example:
<img width="485" alt="image" src="https://user-images.githubusercontent.com/8356424/221436098-a57c4865-beb0-40e0-8dd9-0fd5e5c0f7f2.png">

patriciogonzalezvivo:
Hi @mathiasbredholt this is a great contribution! Do you mind making it into a PR?

mathiasbredholt:
I will!

PR #46, opened 2023-03-05
feat: adding triTile

mathiasbredholt:
Adding a triTile function for creating triangle shaped tiles as suggested in #45 
- Loosely inspired by FabriceNeyret2 hexagonal tiling tuto https://www.shadertoy.com/view/4dKXR3
- The coords are skewed by the rotation matrix used to create the triangles. If might be nice if the output coords weren't skewed so the user can make non-skewed shapes inside the triangles. If anyone has a suggestion how to do that it would be most welcome! 
- I'm not familiar with the hlsl language and haven't got a setup to test it, can you make the translation @patriciogonzalezvivo?

### DIFF
--- a/space/triTile.glsl
+++ b/space/triTile.glsl
@@ -1,0 +1,20 @@
+/*
+original_author: Mathias Bredholt
+description: make some triangular tiles. XY provide coords inside of the tile. ZW provides tile coords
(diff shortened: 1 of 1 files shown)


patriciogonzalezvivo:
This is perfect @mathiasbredholt ! thank you!

PR #50, opened 2023-03-23
fix quotes in bracketing.glsl yaml

crazylafo:
fix quotes around bracketing description
add "n" to orientation


patriciogonzalezvivo:
Thank you!

PR #59, opened 2023-04-08
fix: "sample: Illegal use of reserved word" on WebGL shaders

### LINKED ISSUE #58: Shader Error: "sample: Illegal use of reserved word on WebGL 2.0"
**Issue**: Many Lygia shaders stop working on WebGL 2.0 with this error:

```
sample: Illegal use of reserved word
 ```
 
 **How to reproduce**:  run an [updated version of lygia_threejs_examples](https://github.com/patriciogonzalezvivo/lygia_threejs_examples/compare/main...davcri:lygia_threejs_examples:main).
 
 image
 
 **Hardware**: tested on a MacBook Air with M1 chip.
 
 The issue is not related to ThreeJS as I managed to replicate the error even with a basic WebGL project. Apparently `sample` is a reserved word in WebGL 2.0, even though I didn't manage to find [any reference in the GLSL ES 3.00 spec](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf).

davcri:
Fix #58 

This PR removes `sample()` definition (as I didn't find any usage reference of `FNC_SAMPLE` in the source code), but IDK what's the best choice for Lygia, I only wanted to highlight the issue.

### DIFF
--- a/sample.glsl
+++ b/sample.glsl
@@ -5,8 +5,3 @@
 #define SAMPLER_FNC(TEX, UV) texture2D(TEX, UV)
 #endif
 #endif
-
-#ifndef FNC_SAMPLE
-#define FNC_SAMPLE
-vec4 sample(sampler2D tex, vec2 uv) { return SAMPLER_FNC(tex, uv); }
-#endif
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
Good point. Thank you!

PR #60, opened 2023-04-15
Port random and snoise to WGSL

davidar:
I went looking for a library like this a while ago, glad to see it finally exists :)


patriciogonzalezvivo:
Thank you for contributing and being part of it!

PR #62, opened 2023-05-01
feat(raymarchCast): template distance parameters and better default values

Lutymane:
Hello,

I've added new macros to configure raymarching distance parameters, because the provided fixed values are unsuitable for each case.


patriciogonzalezvivo:
Thank you!

PR #63, opened 2023-06-17
Add required files so that the library can be imported as a Unity package.

kyuuzou:


patriciogonzalezvivo:
Thank you!

kyuuzou:
Hello @patriciogonzalezvivo!

First of all, thank you for creating this library, as well as the Book of Shaders, both are amazing resources. 

About this pull request: I should have communicated more: the files included are actually not enough to have it working as a Unity package, which is I preemptively closed the pull request. I hadn't noticed the .gitignore file was ignoring all meta files, which threw me off (_works on my machine™_). Without them, when Unity tries to import the package, it tries to generate the missing meta files (one per file), and cannot, because the imported package is immutable.

To fix it, all meta files are necessary, which I've tested on this commit ([9615fa6b3c4a64c3b3a5c295ba4ddd66fa16d735](https://github.com/kyuuzou/lygia/commit/9615fa6b3c4a64c3b3a5c295ba4ddd66fa16d735). I'm not sure if you'd want to merge this though, as it will double the amount of files on your repo with Unity-specific stuff?

PR #65, opened 2023-06-27
Add arrow 3D SDF for better debugging

ammein:
I added one SDF for Arrow that can be used for debugging in 3D SDF, Raymarching or etc. I copy pasted the code from shadertoy example and put the original author in the file. Let me know if anything else is needed for this merging.


patriciogonzalezvivo:
Fantastic! Thank you! 
Let's function name and file to `arrowSDF` we could overload the same command with a `vec2` for 2D in the future

ammein:
Done change filename & function name. Is this okay? @patriciogonzalezvivo 🥳

patriciogonzalezvivo:
That's great! thank you!

PR #61, opened 2023-04-29
fix(lighting/raymarch.glsl): typo

Lutymane:

PR #66, opened 2023-07-01
Fakecube fix & typos

ammein:
Resolve custom fakecube uv undeclared error that previously use 'd' variable instead of '_normal'. Also fix the typo issue raised by `Lutymane:patch-1` https://github.com/patriciogonzalezvivo/lygia/pull/61 .

I tried your custom fake cube! LOVE IT!! 🚀🥳 I've been searching for this method for me to do custom fakeCube without using samplerCube/cubemap texture by using `envMap` function!


patriciogonzalezvivo:
Thank you!

PR #70, opened 2023-07-07
Hotfix Raymarch Glass

ammein:
I'm so sorry that I am bothering your day. It seems that my example shader is not loaded properly here: https://lygia.xyz/lighting/raymarch/glass .

So I thought, the format of my example link is not loaded correctly due to the extra character `|` before the file relative link. I also remove some useless code & did refactoring for flexibility. I hope you can merge this with ease! 🥳

### DIFF
--- a/lighting/raymarch/glass.glsl
+++ b/lighting/raymarch/glass.glsl
@@ -1,7 +1,4 @@
 #include "../envMap.glsl"
-#include "../fresnel.glsl"
-#include "../specular.glsl"
-#include "../reflection.glsl"
 
 /*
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
Hi! for the example to work you need to provide a URL to the example. Those with just a relative path are shaders that are in the server where lygia.xyz is hosted. Here is an example: https://github.com/patriciogonzalezvivo/lygia/blob/89ea3897687631dbb0ec89e30b44ea354489b132/math/bump.glsl#L8

Issue #72, opened 2023-07-13
A new integration example


cdaein:
Hi,
I have a creative coding helper package [`create-ssam`](https://github.com/cdaein/create-ssam) that comes with LYGIA + OGL or Three.js templates for a fullscreen shader. Will it be helpful to add the link to the examples in README?

It comes with a few extra helpful features such as video(gif, webm, mp4) exports, git image snapshot for archival, etc. which might be useful for some people.

patriciogonzalezvivo:
That would be great! Thank you!

PR #73, opened 2023-07-13
Update README.md to add 2d integration examples

cdaein:
from issue #72:

added 2d integration examples (OGL, Three.js) using [Ssam/create-ssam](https://github.com/cdaein/create-ssam) packages.


patriciogonzalezvivo:
Fantastic! Thank you!

PR #74, opened 2023-07-14
Fix broken include paths

MoltenCoffee:
Hi,
This small PR fixes broken includes in aabb.glsl and aabb.hlsl (`aabb/center.*` instead of `aabb/centroid.*`), as well as a typo in the `use` field of the documentation of `centroid.hlsl` (`centrood` instead of `centroid`).

Regards, TJ


patriciogonzalezvivo:
Thank you!

PR #78, opened 2023-07-28
Fix 1D random use of RANDOM_SINLESS

JeffreyPalmer:
Was reading through the code today and noticed that the RANDOM_SINLESS logic for 1D random was inverted. This PR fixes that inconsistency.

Thanks for the amazing library! 🙏


patriciogonzalezvivo:
thank you! good catch!

PR #79, opened 2023-07-28
Add mandelbulbSDF and spherical coord functions

kfahn22:
This adds the SDF for the mandelbulbSDF and 3D spherical coordinates (which is also used for 3D supershape).

FYI, this is the sketch (slightly updated to use mandelbulbSDF in pull request)  I did previously.  Most of ray marching functions are ported from watch Martijn Steinrucken.  I modeled the SDF partly on  Daniel Shiffman's mandelbulb coding challenge (which created a point cloud)  and is also based on Inigo Quilez's version of the mandelbulb.

I should add that the SDF returns a vec2 and raymarching  is more complicated than for others, so including an example would be useful.
https://editor.p5js.org/kfahn/sketches/rF-g-OBH0


space/spherical.glsl
```diff
@@ -0,0 +1,17 @@
+/*
+author: Kathy McGuiness
+description: Returns the spherical coordinates
+use: sphericalSDF(<vec3> st)
```
patriciogonzalezvivo, line 4: use: spherical(<vec3> st)

space/spherical.glsl
```diff
@@ -0,0 +1,17 @@
...
+*/
+
+#ifndef FNC_SPHERICAL
+#define FNC_SPHERICAL
+vec3 spherical( in vec3 st)
```
patriciogonzalezvivo, line 9: this is spherical to cartesian? should be good to specify from what to what type of coordenates

kfahn22: yes, will edit

sdf/mandelbulbSDF.glsl
```diff
@@ -0,0 +1,39 @@
+#include "../space/spherical.glsl"
+
+/*
+author: Kathy McGuiness
```
patriciogonzalezvivo, line 4: this is headers are structured in YAML format to compose the documentation in lygia.xyz/sdf. the field for author is: `original author`

sdf/mandelbulbSDF.glsl
```diff
@@ -0,0 +1,39 @@
...
+description: Returns the mandelbulb SDF
+use: mandelbulbSDF(<vec2> st)
+*/
+
+#ifndef FNC_MANDELBULBSDF
+#define FNC_MANDELBULBSDF
+vec2 mandelbulbSDF( in vec3 st ) 
+{
+  vec3 zeta = st;
```
patriciogonzalezvivo, line 13: 4 spaces as tabs

space/spherical.glsl
patriciogonzalezvivo: let's re use cart2polar but overload it so together with the `vec2 cart2polar(vec2)` we have this one with `vec3 cart2polar(vec3)`

patriciogonzalezvivo:
Thank you again for this new cool features

sdf/mandelbulbSDF.glsl
kfahn22: OK, will fix.  VSC wasn't automatically formatting for me.  Do you have an extension you use?

patriciogonzalezvivo: I don't : ( sorry.
I just set the tabs as 4 spaces.

patriciogonzalezvivo:
I did some editions to the styling on your previous commit. Hope you don't mind. I thought it may be useful. 
https://github.com/patriciogonzalezvivo/lygia/commit/3ec3024aaf156d9db347178bc0ef927a5926fe33

kfahn22:
I am fine with whatever changes you make.  I just made some updates (updated name of spherical function) and tried to fix spacing, and just hope that I didn't create conflicts.

patriciogonzalezvivo:
Sorry here https://github.com/patriciogonzalezvivo/lygia/commit/66b52b7f78bc28eeb1a1b04c3791ccc7475c4959

space/cart2polarXYZ.glsl
```diff
@@ -0,0 +1,16 @@
+/*
+description: Returns the cartesian to polar coordinates
+use: spherical(<vec3> st)
+*/
+
+#ifndef FNC_CART2POLARXYZ
+#define FNC_CART2POLARXYZ
+vec3 cart2polarXYZ( in vec3 st )
```
patriciogonzalezvivo, line 8: sorry maybe I didn't express correctly. **Overloading** means that have the same exact name but the arguments are different. In this case this fuction will have the same `cart2polar` name in the `cart2polar.glsl` but will be different because they have different argument types. Like for example: https://github.com/patriciogonzalezvivo/lygia/blob/main/math/bump.glsl#L14

sdf/mandelbulbSDF.hlsl
```diff
@@ -1,37 +1,36 @@
...
-  for (int i = 0; i < maxiterations; i+=1) {
-     dz = n*pow(m, 3.5)*dz + 1.0;
-     vec3 sphericalZ = spherical( zeta ); 
-     float newx = pow(sphericalZ.x, n) * sin(sphericalZ.y*n) * cos(sphericalZ.z*n);
-     float newy = pow(sphericalZ.x, n) * sin(sphericalZ.y*n) * sin(sphericalZ.z*n);
-     float newz = pow(sphericalZ.x, n) * cos(sphericalZ.y*n);
-     zeta.x = newx + st.x;
-     zeta.y = newy + st.y;
-     zeta.z = newz + st.z;
+vec2 mandelbulbSDF( in vec3 st ) {
```
patriciogonzalezvivo, line 11: Make sure when you make the HLSL version to use HLSL types and functions. vec2 are call float2 for example

kfahn22: Sorry I wan't aware of that!

space/cart2polarXYZ.glsl
kfahn22: Sorry, I don't have a lot of experience with overloading functions.  So instead of adding new functions I should edit the existing files to add new version, correct?

kfahn22:
I have added 3d version to original car2polar files and updated the hlsl files (hopefully correctly).  Sorry for my mistakes!

space/cart2polar.glsl
```diff
@@ -9,4 +9,10 @@ use: cart2polar(<vec2> st)
 vec2 cart2polar(in vec2 st) {
     return vec2(atan(st.y, st.x), length(st));
 }
+vec3 cart2polar( in vec3 st ) {
```
patriciogonzalezvivo, line 12: perfect!

space/cart2polar.hlsl
```diff
@@ -9,4 +9,11 @@ use: cart2polar(<float2> st)
 float2 cart2polar(in float2 st) {
     return float2(atan2(st.y, st.x), length(st));
 }
+float3 cart2polar( in float2 st)
```
patriciogonzalezvivo, line 12: here the argument type should be a `float3`

patriciogonzalezvivo:
Perfect! thank you! 
I added an example for the supershape here https://github.com/patriciogonzalezvivo/lygia_examples/commit/4b47a0bcbc6d01ea726642b7e523ea9786e85c0a 
Do you mind creating an example for this one?

kfahn22:
Yes, I will try to create an example.

kfahn22:
I have been working on creating an example and I don’t think the render is going to be very good unless there is a special ray marching function just for the mandelbulb because the number of iterations is critical to the render.   I am also concerned that my version doesn’t render fast enough with it animating.   I am not sure that is something you will want to do, so would it be best to remove the mandelbulbSDF?  Again, I am very sorry that did not occur to me  to check this out ahead of time before submitting a pull request and wasting your time!!!  (I also realized that I forgot to update the name of the spherical function in the mandelbulbSDF, but that will not matter if it is deleted.). Let me know and I can create a pull request removing.

 This is version with ray marching returning the number of iterations taken & using lygia coloring (which is OK, but there are certainly far better versions):

https://editor.p5js.org/kfahn/sketches/SnfJTVFrV￼

This is version using default lygia ray marching :

https://editor.p5js.org/kfahn/sketches/-mv_Dblcy￼

PR #80, opened 2023-07-29
Add 3d supershape, update mandelbulb

kfahn22:
I realized when I started working on an example for the mandelbulb that I neglected to update the name of the function used to get the spherical coordinates so I have fixed that.   I also realized that I used the raymarch f'n to return iterations and that won't work with your code, so I am working on figuring out how to color the mandelbulb using your functions.  I apologize--it has been a while since I looked at the mandelbulb code and I had forgotten how complicated it is.

I have also added the 3d version of the supershape.  (I wasn't quite sure what I should name it.)  I have tried to add a bit of information about good parameters and a link to a website that has some amazing versions done in Processing.  My version is not as complex ( it could easily be edited, but that would double the already long list of parameters.). 
Here is the 3d supershape sketch with some parameters.   I can add an example for this.
https://editor.p5js.org/kfahn/sketches/4D5mvO_BN

I added the superShapeSDF, mandelbulbSDF, and superShape3dSDF to the list of SDFS.


patriciogonzalezvivo:
Hi! I saw you close this PR. Was that intentional?

patriciogonzalezvivo:
That looks super cool! thanks again for all the time and dedication. Loving all you are bringing to LYGIA

PR #88, opened 2023-08-03
Add koch sdf

kfahn22:
While this  is not an exact copy of The Art of Code's version of the Koch curve,  it is essentially his code.  I have added if statements to choose different iterations.  I couldn't find a function that returned vec2(sin(angle), cos(angle)).  Please let me know if there is one that can be utilized.

### DIFF
--- a/sdf/kochSDF.glsl
+++ b/sdf/kochSDF.glsl
@@ -1,0 +1,50 @@
+#include "../math/const.glsl"
+
+/*
+original_author: The Art of Code
+description: Returns a Koch curve SDF    
+use: <vec2> kochSDF(<vec2> st, <int> iterations)
+*/
+
+#ifndef FNC_KOCHSDF
+#define FNC_KOCHSDF
+vec2 kochSDF( vec2 st, int N ) {
+    float angle = (5.0/6.0)*PI;
+    st = st * 2.0 - 1.0;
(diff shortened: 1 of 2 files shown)


kfahn22:
BTW, I am playing around with the code in morphological and get this error trying to use alphaFill (which is why I added if statements to the koch curve code.)  I am using p5.js.

's' : Loop index cannot be compared with non-constant expression

patriciogonzalezvivo:
Have you add the following line to the top of your code?
```
#define PLATFORM_WEBGL
```

There is no dynamic arrays on WebGL 1.0, that flag before importing any function, changes the code to be back compatible with WebGL1.0

sdf/kochSDF.glsl
```diff
@@ -0,0 +1,50 @@
+#include "../math/const.glsl"
+
+/*
+original_author: The Art of Code
```
patriciogonzalezvivo, line 4: maybe we should put the name of the person instead of his youtube channel?

kfahn22: I agree.  I used Art of Code because another file referred to the original author that way.

kfahn22:
I followed the pattern for alphaFill to add loop-please let me know if I need to fix.

PR #90, opened 2023-08-06
ADD/sample/marchingSquares.glsl: first draft of marching squares algorithm

guidoschmidt:
Hey @patriciogonzalezvivo I've added a first version of a marching squares algorithm. Would love to get some feedback on naming, code improvements, potential performance improvements and everything else that comes to your mind ✌🏽 

I've also put up an example on how to use it over at my [examples.lygia-threejs](https://github.com/guidoschmidt/examples.lygia-threejs) repo.

### DIFF
--- a/sample/marchingSquares.glsl
+++ b/sample/marchingSquares.glsl
@@ -8,6 +8,9 @@ description: |
     The Marching Squares algorithm generates contour lines (isolines) of a given input for a
     given grid and a threshhold. Is uses a lookup of 16 cases to decide if a cell is inside, outside or
     a specific corner.
+    **References:**
+    - [urbanspr1nter.github.io](https://urbanspr1nter.github.io/marchingsquares/)
+    - []()
 use: <vec3> sampleMarchingSquares(in <vec2> uv, in <sampler2D> tex, in <float> cellSize, in <float> threshold, in <vec2> resolution) 
 options:
     - FNC_SAMPLEMARCHINGSQUARES(TEX, UV): optional sampling function
@@ -24,7 +27,7 @@ float defaultSampleFunction(in sampler2D tex, in vec2 uv) {
 
 #define centerOf(p1, p2) mix(p1, p2, 0.5)
 
-float sampleOutline(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in bool straight) {
+float sampleMarchingSquares_outline(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in bool straight) {
     float lineStrength = (straight ? 1.3333 : 1.0);
     vec2 pa = p - a;
     vec2 ba = b - a;
@@ -32,7 +35,7 @@ float sampleOutline(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in bool str
     return 1.0 - step(lineStrength, length(line));
 }
 
-float sampleTile(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in int tile) {
+float sampleMarchingSquares_tile(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in int tile) {
     vec2 pa = p - a;
     vec2 ba = b - a;
     vec2 line = pa - ba * dot(pa, ba) / dot(ba, ba);
@@ -71,8 +74,6 @@ float sampleTile(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in int tile) {
           return shape13_0 - shape13_1;
         default:
           return 0.0;
-        // @DEBUG: enable the smoothstep in order to show a line for tiles
-        //return smoothstep(0.4, 0.0, length(line));
     }
 }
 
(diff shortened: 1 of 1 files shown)


sample/marchingSquares.glsl
```diff
@@ -0,0 +1,204 @@
+#include "../generative/pnoise.glsl"
+#include "../math/const.glsl"
+#include "../math/rotate2d.glsl"
+#include "../sample.glsl"
+
+/*
+Author: Guido Schmidt
```
patriciogonzalezvivo, line 7: the field should be `original_author:  Guido Schmidt`

sample/marchingSquares.glsl
```diff
@@ -0,0 +1,204 @@
...
+*/
+
+#ifndef SAMPLE_MARCHING_SQUARES_FNC
+float defaultSampleFunction(in vec2 uv) {
+    float dv = pnoise(uv * 5.0, vec2(0.0));
+    return dv;
+}
+#define SAMPLE_MARCHING_SQUARES_FNC(UV) defaultSampleFunction(UV)
+#endif
+
```
patriciogonzalezvivo, line 23: We use the following pattern to prevent name collisions

```
#ifndef FNC_SAMPLEMARCHINGSQUARES 
#define FNC_SAMPLEMARCHINGSQUARES
...
#endif
```

sample/marchingSquares.glsl
```diff
@@ -0,0 +1,204 @@
...
+Author: Guido Schmidt
+description: The Marching Squares algorithm generates contour lines (isolines) of a given input for a
+given grid and a threshhold. Is uses a lookup of 16 cases to decide if a cell is inside, outside or
+a specific corner.
+use: <vec3> marchingSquares(in <vec2> uv) 
+options:
+    - SAMPLE_MARCHING_SQUARES_FNC(UV): optional sampling function, defaults so sampling of pnoise
+*/
+
+#ifndef SAMPLE_MARCHING_SQUARES_FNC
```
patriciogonzalezvivo, line 16: the name of this macros should be

```
SAMPLEMARCHINGSQUARES_FNC(TEX, UV)
```

I see no sign of texture sample... would be that possible to add?
We could have to main sampleMarchingSquares functions one that takes a texture and one that doesn't

sample/marchingSquares.glsl
```diff
@@ -0,0 +1,204 @@
...
+          float shape13_1 = step(1.05, (rotate2d(PI * 0.25) * cellUv).x);
+          return shape13_0 - shape13_1;
+        default:
+          return 0.0;
+        // @DEBUG: enable the smoothstep in order to show a line for tiles
+        //return smoothstep(0.4, 0.0, length(line));
+    }
+}
+
+vec2 marchinSquares(in vec2 uv, in float cellSize, in float threshold, in vec2 resolution) {
```
patriciogonzalezvivo, line 78: functions inside `sample` folder start with `sample` in this case would be `sampleMarchingSquares()`

sample/marchingSquares.glsl
```diff
@@ -0,0 +1,204 @@
...
+float defaultSampleFunction(in vec2 uv) {
+    float dv = pnoise(uv * 5.0, vec2(0.0));
+    return dv;
+}
+#define SAMPLE_MARCHING_SQUARES_FNC(UV) defaultSampleFunction(UV)
+#endif
+
+#define centerOf(p1, p2, v1, v2) mix(p1, p2, 0.5)
+
+float sampleOutline(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in bool straight) {
```
patriciogonzalezvivo, line 26: this can potentially produce name collisions. 
Let's put the name of the main function first like
```
float sampleMarchingSquares_outline(...) {
}
```

sample/marchingSquares.glsl
```diff
@@ -0,0 +1,204 @@
...
+    float lineStrength = (straight ? 1.3333 : 1.0);
+    vec2 pa = p - a;
+    vec2 ba = b - a;
+    vec2 line = pa - ba * dot(pa, ba) / dot(ba, ba);
+    return 1.0 - step(lineStrength, length(line));
+}
+
+float sampleTile(in vec2 p, in vec2 cellUv, in vec2 a, in vec2 b, in int tile) {
```
patriciogonzalezvivo, line 34: Same here

sample/marchingSquares.glsl
```diff
@@ -0,0 +1,204 @@
...
+
+#ifndef SAMPLE_MARCHING_SQUARES_FNC
+float defaultSampleFunction(in vec2 uv) {
+    float dv = pnoise(uv * 5.0, vec2(0.0));
+    return dv;
+}
+#define SAMPLE_MARCHING_SQUARES_FNC(UV) defaultSampleFunction(UV)
+#endif
+
+#define centerOf(p1, p2, v1, v2) mix(p1, p2, 0.5)
```
patriciogonzalezvivo, line 24: I don't really get this? why it takes `v1` and `v2` ??

patriciogonzalezvivo (changes requested):
Hi Guido! This is a fantastic addition.

Not 100% if `sample/` is the right place.
I like that you can sample noise fields, would make sense in this folder if it was set up to sample a texture and in the absence of it a noise filed. 

What do you think?

Other wise maybe it's place could be on `sdf/`? `geometry/`? `draw/`? Mmm..... I'm really not sure.

sample/marchingSquares.glsl
guidoschmidt: Oh yes, I struggled with this, too. I've went with `sample` because I felt the algorithm is kind of resampling a given input to create isolines. `sdf` could also be a good option (though I'm using it for non-sdf related things, too) 🤔

sample/marchingSquares.glsl
guidoschmidt: Oh yes, I wanted to add the possibility to sample from a texture

sample/marchingSquares.glsl
patriciogonzalezvivo: it's a tricky one, right? I think by sampling a texture by default we make it have sense here. I'm glad you were also thinking on that and I'm not pushing you in any direction you don't want to go

guidoschmidt: Oh right, I was taking the `sample/2DCube` as a reference, it has `Author`, will change 👍🏽

patriciogonzalezvivo: Oh! you are right! Good catch! My bad, I need to fix it https://github.com/patriciogonzalezvivo/lygia/blob/main/sample/3DSdf.glsl#L7 
I love that I put capital `A` like an egomaniac lol

guidoschmidt:
Alright, just had a quick catch on testing it using a sampled texture, and it seems like there are some artifacts (eventually due to mismatch of texture and screen resolution? 🤔 ). 

Anyway, tried to integrate all of the other requested changes . Also is there a potential field like `references` to point to some helpful resources?

patriciogonzalezvivo:
Lovely! excited for! there is no reference. I usually added it to a multi-line on `description: | ` that gets parse as markdown for the online documentation. Ex:

https://github.com/patriciogonzalezvivo/lygia/blob/main/color/mixSpectral.glsl#L8-L12

https://lygia.xyz/color/mixSpectral

sample/marchingSquares.glsl
patriciogonzalezvivo: Sorry this comment I made was a bit confusing. 

```
#ifndef FNC_SAMPLEMARCHINGSQUARES 
#define FNC_SAMPLEMARCHINGSQUARES
...
#endif
```

is to wrap the entire function. For example here: https://github.com/patriciogonzalezvivo/lygia/blob/main/sample/shadowPCF.glsl#L18-L19

While for "templeting" just the function use to do sampling have this other pattern:

```
#ifdef SAMPLEMARCHINGSQUARES_SAMPLE_FNC
#define SAMPLEMARCHINGSQUARES_SAMPLE_FNC(TEX, UV) defaultSampleFunction(TEX, UV)
#endif
```

For example:

sample/marchingSquares.glsl
```diff
@@ -0,0 +1,205 @@
...
+          float shape13_1 = step(1.05, (rotate2d(PI * 0.25) * cellUv).x);
+          return shape13_0 - shape13_1;
+        default:
+          return 0.0;
+        // @DEBUG: enable the smoothstep in order to show a line for tiles
+        //return smoothstep(0.4, 0.0, length(line));
+    }
+}
+
+vec2 sampleMarchinSquares(in vec2 uv, in sampler2D tex, in float cellSize, in float threshold, in vec2 resolution) {
```
patriciogonzalezvivo, line 79: We usually sort the arguments in this orders: `sample2D`, `vec4`, `vec3`, `vec2`, `float`, `int`

patriciogonzalezvivo:
Hi Guido! Thanks for the changes! I add a couple more notes. Hope you don't mind.
Thanks again

sample/marchingSquares.glsl
guidoschmidt: oh yes that's actually a residue, good catch

sample/marchingSquares.glsl
guidoschmidt: Ah yes, got it. Really like in C++ world with the header files, thanks for clarifying

sample/marchingSquares.glsl
guidoschmidt: Ah nice, I see. I like this idea of ordering

PR #91, opened 2023-08-06
Replace one value for c and add gist

kfahn22:
It looks lovely with the spectral color:
Screenshot 2023-08-06 at 12 23 24 PM

### DIFF
--- a/sdf/juliaSDF.glsl
+++ b/sdf/juliaSDF.glsl
@@ -7,10 +7,12 @@ description: |
     Some values for c:
         * vec2(−0.8, 0.156)
         * vec2(0.285, 0.0)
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
Nice! Question, do you have twitter? a website? I would love to share the contributions you are doing and give you credit

kfahn22:
I do have a have a twitter,  but never use it.  I have started a website,
but still a work-in-progress.

 I am happy to contribute —this is such a great resource!

On a side note, after  thinking about it I am wondering whether  adding a
separate folder for fractals might be a good idea.    Is that something you
would be willing to consider?  (I am thinking we could put julia set there
and add Sierpinski’s triangle as well as other fractal code.)

PR #94, opened 2023-08-11
Increase size of returned triangle

kfahn22:
When rendering a triangle, I found that it was very small  (maybe just me?)  so I have increased the size of the triangle.

### DIFF
--- a/sdf/triSDF.glsl
+++ b/sdf/triSDF.glsl
@@ -6,8 +6,9 @@ use: triSDF(<vec2> st)
 
 #ifndef FNC_TRISDF
 #define FNC_TRISDF
-float triSDF(in vec2 st) {
-    st = (st * 2. - 1.) * 2.;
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
good catch

patriciogonzalezvivo:
Sorry, I revert this change. 2D shapes are not really SDFs, the 0 is not in the surface but in 0.5. 
Also they are aligned to 0.5,0.5 as the center unless CENTER_2D is define as something different
Sorry for the inconvenience, this is to keep them consistent to each other
https://github.com/patriciogonzalezvivo/lygia_examples/blob/main/draw_shapes.frag

patriciogonzalezvivo:
I use the oportunity to fix all of them in this commit https://github.com/patriciogonzalezvivo/lygia/commit/456efe89cea9f4cdc3646998b645e6374246779e

PR #95, opened 2023-08-13
Fixing some issues

patriciogonzalezvivo:
GLSL 120 don't support switch


patriciogonzalezvivo:
@guidoschmidt what do you think? hope is ok the changes

PR #101, opened 2023-10-11
removed unterminated define in rgb2yiq.hlsl

pajama:
I'm starting to go through the old unity samples and update them to the latest lygia. There are a few hlsl errors that are failing to compile. Some have me stumped, but I'll try to push up the simple ones like this as I get through them.


patriciogonzalezvivo:
Thank you

PR #105, opened 2023-10-12
some more hlsl fixes for unity

pajama:


patriciogonzalezvivo:
You are killing it! Thanks again

PR #109, opened 2023-10-22
new blending functions: hue, sat, lum, color

couleurs:
Adds 4 blending functions that are common in applications like Photoshop or Figma:
- Hue blend
- Saturation blend
- Luminosity blend
- Color blend


patriciogonzalezvivo:
Gracias amigo

PR #111, opened 2023-11-08
Request to add a kaleidoscopic space transformation 

marioecg:


patriciogonzalezvivo:
@marioecg this is perfect! wonder if the overloading functions with less argument can be just calls to the third (and longer) but with the defaults. So there is less repeating code. Beside that this is absolutely perfect

space/kaleidoscope.glsl
```diff
@@ -0,0 +1,60 @@
+#include "../math/const.glsl"
+
+/*
+contributors: [Mario Carrillo, Daniel Ilett]
+description:
```
patriciogonzalezvivo, line 5: You need to add `|` before doing multi-line. That's a YAML thing

```
description: |
```

space/kaleidoscope.glsl
```diff
@@ -0,0 +1,60 @@
...
+    Converts carteesian coordinates into polar coordinates 
+    emulating a kaleidoscope visual effect.
+    Based on Daniel Ilett's tutorial on reflecting polar coordinates: 
+        https://danielilett.com/2020-02-19-tut3-8-crazy-kaleidoscopes/
+*/
+
+// Default use when just the coordinates are given, the segment count is set to 8
+vec2 kaleidoscope(in vec2 coord) {
```
patriciogonzalezvivo, line 13: Something like:

```
vec2 kaleidoscope(in vec2 coord) {
    return kaleidoscope(coord, 8, 0.0) {   
}

vec2 kaleidoscope(in vec2 coord, in float segmentCount) {
    return kaleidoscope(coord, segmentCount, 0.0) {   
}


```

space/kaleidoscope.glsl
```diff
@@ -0,0 +1,60 @@
...
+    float segmentCount = 8.0;
+    
+    vec2 uv = coord - 0.5;
```
patriciogonzalezvivo, line 16: Consider using `CENTER_2D` like here https://github.com/patriciogonzalezvivo/lygia/blob/main/space/scale.glsl#L14

marioecg:
@patriciogonzalezvivo oh you're right! I'll be pushing all those changes in short. Could you check that the change for `CENTER_2D` is correct? I am a little uncertain if that's how you should use it like.

patriciogonzalezvivo:
Beautiful!

patriciogonzalezvivo:
thank you!

PR #116, opened 2023-11-11
Update kaleidoscope.glsl

edap:
Replace st with kuv, use float instead of int


marioecg:
@edap thanks for catching this! I just spotted it while trying to use the function.

PR #129, opened 2024-03-02
Fix broken Oklab color space matrices and conversions.

patternspandemic:
Hello.

This PR resolves what seems to be a big of mix up with the Oklab color space functions, naming, and its use of the srgb <--> (linear)rgb conversion functions.

I think transformation functionality was misnamed and/or put into the wrong files in the two different conversion scenarios: oklab2(s)rgb and (s)rgb2oklab. This problem was then exacerbated by using the wrong order of linear (rgb) and non-linear (srgb) transformations, probably due to the naming and placement of functionality.
For example,
* the current _oklab2srgb_ file actually converts oklab to (linear)rgb, but is named as converting to sRGB. 
* the current _oklab2rgb_ file uses the incorrectly named `oklab2srgb` to perform a correct transform of oklab->linear, but then incorrectly uses `srgb2rgb` (a double transform to linear).
* the current _rgb2oklab_ converts (linear)rgb to sRGB (even though its already linear) probably because the actual transform functionality was put into _srgb2okab_.
* the current _srgb2oklab_ actually converts (linear)rgb to oklab, but is named as converting from sRGB.

For testing, I used [this pinkish color](https://oklch.com/#72.24267371855285,0.2068369609861689,350.13962091155344,100) and the srgb and oklab derived values provided by the color picker functionality. I could not achieve results that were expected with any combination of current functions. These incorrect results can be seen by using this [Oklab tester shader](https://gist.github.com/patternspandemic/720cbee05d4564c770721978e0442d9a) which I creatred to confirm this fix. With existing code, the color turned out on the lime green side of things rather than pinkish. With this fix, the expected results are a 2x2 grid of the same pinkish color.

I think there may have been one or two errors in the transform matrices involved. I have replaced them with the current ones found at the resource here, which I utilized to perform this fix:
[Converting from linear sRGB to Oklab](https://bottosson.github.io/posts/oklab/#converting-from-linear-srgb-to-oklab)
[Use a linear version of sRGB](https://bottosson.github.io/posts/colorwrong/#what-can-we-do%3F)

I can only confirm the glsl functionality works as I expect, and was unable to test the updated hlsh/wgsl code.

Cheers, Brad


patriciogonzalezvivo:
Thank you so much!

PR #133, opened 2024-04-20
fix some hlsl include to glsl in triangle.glsl and make_vec3 to vec3

jane00:
fix some hlsl include to glsl in triangle.glsl and make_vec3 to vec3 in barycentric.glsl


patriciogonzalezvivo:
Thank you!

Issue #130, opened 2024-03-25
"smooth" reserved keyword


kylegrover:
"smooth" is a reserved keyword in GLSL
https://github.com/patriciogonzalezvivo/lygia/blob/b364f4d6481450ce4aab9a0dac86ad8dc8089125/lighting/pbrLittle.glsl#L58
https://github.com/patriciogonzalezvivo/lygia/blob/b364f4d6481450ce4aab9a0dac86ad8dc8089125/lighting/pbrLittle.glsl#L75

patriciogonzalezvivo:
good catch! Do you mind making a PR?

UstymUkhman:
@kylegrover I hope you don't mind I've created a PR for this 'cause I might need that shader soon. :sweat_smile:

PR #136, opened 2024-04-28
Fix 'smooth' reserved keyword #130

UstymUkhman:
I've also updated the `.hlsl` file to keep shaders as similar as possible.


patriciogonzalezvivo (approved):

patriciogonzalezvivo:
Thank you and welcome to the comercial license https://lygia.xyz/license

PR #141, opened 2024-06-24
Improve color mixing for SDF smooth union and subtraction operations

bryndsey:
Previously, when performing union or subtraction operations with smoothing of 2 SDFs that used the same color, the operation would change the color (i.e. making it lighter for subtraction or darker for union). The amount of color variation increased as the value for the operation smoothing increased. See these screenshots for example, where the skin color should be uniform (discounting variations due to lighting):

| Small smoothing value | Larger smoothing value |
| ------------- | ------------- |
| image | image |

Modifying the union and subtraction operation functions as in this PR seems to resolve the color mixing, and produces the expected result:

image


bryndsey:
Note: I haven't tested any changes with anything other than GLSL. I updated the others for consistency, but I don't have a setup to test it.

patriciogonzalezvivo:
Thank you!

PR #143, opened 2024-06-27
Fixed typo bugs in hlsl versions of yiq2rgb and hsv2rgb

shadielhajj:
Both functions were broken because of typos in variable names.


patriciogonzalezvivo:
Thank you

PR #146, opened 2024-06-29
Fixed pbrLittle indirect diffuse.

shadielhajj:
Fixed lighting/pbrLittle. The indirect diffuse component (spherical harmonics) should be added to direct diffuse, not multiplied by it.

pbr little fix


patriciogonzalezvivo:
Nice catch! Thank you

PR #147, opened 2024-07-01
Aligned HLSL version of lighting/specular/cookTorrance with GLSL version

shadielhajj:
Added pluggable NDF support (with GGX as default) to lighting/specular/cookTorrance,  bringing it in line with GLSL version.


patriciogonzalezvivo:
Thank you!

PR #150, opened 2024-07-04
Roughness-adjusted fresnel.

shadielhajj:
Added a roughness-adjusted fresnel function. This function has been described by Sébastien Lagarde and others.

The rationale is as follows: "Applying Fresnel term to prefiltered cubemap has a bad effect of always showing high specular color at edge, even for rough surface.... We approximate this “fresnel attenuation” empirically by introducing a fudge factor (taking count of gloss) in fresnel equation to get pleasant visual result"

This produces significant improvements when applied to the indirect components of a PBR shader.

roughness fresnel

For more details, see Lagarde's blog post.
https://seblagarde.wordpress.com/2011/08/17/hello-world/

### DIFF
--- a/lighting/fresnel.glsl
+++ b/lighting/fresnel.glsl
@@ -1,5 +1,6 @@
 #include "common/schlick.glsl"
 
+#include "../math/pow5.hlsl"
 #include "../math/const.glsl"
 #include "../math/saturate.glsl"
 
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
Nice improvement! Thank you!

PR #151, opened 2024-07-04
Complete overhaul of the HLSL PBR shader to align with the GLSL Version.

shadielhajj:
Complete overhaul of the HLSL PBR shader to align with the GLSL Version. This includes the following dependencies:
* Ported LightPoint struct to point.hlsl
* Ported LightDIrection struct to directional.hlsl
* Ported cache fields to material.hlsl
* Ported support for Material struct in point.hlsl, directional.hlsl, fresnelReflection.hlsl and envMap.hlsl
* Ported support for Material struct in specularAO and updated signature to match glsl
* Ported new.hlsl
* Ported resolve.hlsl
* Ported and updated envBRDFApprox.hlsl
* Finally ported pbr.hlsl from glsl

### DIFF
--- a/lighting/camera.glsl
+++ b/lighting/camera.glsl
@@ -1,0 +1,21 @@
+/*
+contributors: Patricio Gonzalez Vivo
+description: Generic Camera Structure
+license:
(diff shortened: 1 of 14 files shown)


patriciogonzalezvivo:
Wow! Massive!

PR #152, opened 2024-07-06
PBR fixes and visual tweaks.

shadielhajj:
Some fixes and visual tweaks to the PBR shader.
I back-to-backed the HLSL version against the Unity Standard Shader and my own PBR implementation (based on the Karis UE4 variant).

1. Fixed the Unity light direction in the HLSL version. It was flipped in PR #151 (my bad!)

2. Tweaked the indirect specular component (glsl and hlsl). I'm not sure what was the rationale for the previous implementation. The new one conforms to the Brian Karis article (https://www.unrealengine.com/en-US/blog/physically-based-shading-on-mobile) and in any case looks much better.
pbr fix

3. As it's currently implemented, the contribution of point and directional light (direct specular and direct diffuse) doesn't look correct to me. It is missing a multiplication by `dot(N, L)`. I added that to point and directional lights, in glsl and hlsl. Again, the new version looks much better and avoids the overblown fresnel.
pbr light fix

### DIFF
--- a/lighting/light/directional.glsl
+++ b/lighting/light/directional.glsl
@@ -39,8 +39,8 @@ void lightDirectional(
     float dif = diffuse(_Ld, _N, _V, _NoV, _NoL, _roughness);
     float spec = specular(_Ld, _N, _V, _NoV, _NoL, _roughness, _f0);
 
-    _diffuse  += max(vec3(0.0), _Li * (_diffuseColor * _Lc * dif));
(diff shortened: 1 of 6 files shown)


patriciogonzalezvivo:
Thank you so much for taking the time. The references really shows the difference. Thanks again for you fantastic contribution

patriciogonzalezvivo:
Do you have a twitter account? I would like to share all the improvements you have been doing

shadielhajj:
Thank you @patriciogonzalezvivo for this fantastic and epic project!

PR #156, opened 2024-07-07
fcos(x): band-limited cos(x), GLSL and HLSL

shadielhajj:
Added IQ's band-limited variant of `cos(x)` which dramatically reduces aliasing at high frequencies.
Copied as is from Inigo's 2020 article https://iquilezles.org/articles/bandlimiting/

`cos(x)` wave aliasing at high frequencies
Screenshot 2024-07-07 140402

`fcos(x)` replacement
Screenshot 2024-07-07 140947

### DIFF
--- a/math/fcos.glsl
+++ b/math/fcos.glsl
@@ -9,8 +9,7 @@ use: fcos(<float> value)
 #ifndef FNC_FCOS
 #define FNC_FCOS
 
-float fcos(in float x)
-{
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
Super useful!

shadielhajj:
Indeed! Love these band-limited variants of step, fract, cos... by IQ and Neyret. I use them by default.

PR #157, opened 2024-07-07
Fixed color space conversion in HLSL

shadielhajj:
Missing `static` keyword was leading to `const` color conversion matrices being optimised away by the hlsl compiler.
Shaders were failing silently and returning black.


patriciogonzalezvivo:
Nice find! So good that you have deeper experience with HLSL

shadielhajj:
Not the first time I get bitten by this one. Not sure if it's a Unity or HLSL thing, but it's horrible because it just fails silently.

PR #159, opened 2024-07-08
Added more missing static keywords (HLSL only).

shadielhajj:
The sequel to #157 All `const` variables are now `static const`, preventing them from being discarded by the HLSL compiler.


patriciogonzalezvivo:
Thank you!

PR #160, opened 2024-07-09
Band-limited aafract (2nd approach)

shadielhajj:
Another take at band-limiting aafract with a slightly different approach. This time I kept the existing algorithm but tacked a low-pass filter to it, centered on the Nyquist frequency (half a pixel).

Summary of changes:
* Added nyquist.glsl and nyquist.hlsl (low-pass filters)
* Calling nyquist() in aafract (+some refactoring of the unnecessary map() call)
* Using the L2 norm instead of L1 (width) in aafract.hlsl, in line with aafract.glsl

Without band-limiting:
Screenshot 2024-07-07 223623

With bandlimiting:
Screenshot 2024-07-09 195048

### DIFF
--- a/math/aafract.glsl
+++ b/math/aafract.glsl
@@ -1,4 +1,4 @@
-#include "map.glsl"
+#include "nyquist.glsl"
 
 /*
 contributors: dahart (https://www.shadertoy.com/user/dahart)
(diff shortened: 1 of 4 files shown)


patriciogonzalezvivo:
brilliant! Thanks again!

patriciogonzalezvivo:
question is the `nyquist` the best way to describe/name the function?

shadielhajj:
@patriciogonzalezvivo It's a low-pass filter really, so I also considered calling it simply `lowpass`. But since it's centered around the Nyquist frequency by default, I finally went for Nyquist. I don't mind changing it. Let me know if you have a preference.

PR #161, opened 2024-07-09
Added distance metrics and Worley distance option (HLSL, GLSL)

shadielhajj:
Added some commonly used distances to math/distance:
* Euclidean
* Manhattan
* Chebychev
* Minkowski (a generalisation of the above three, which takes the DIST_MINKOWSKI_P option)

generative/worley now takes a DIST_FNC option to switch between these distances. euclidean is the default, so the current behaviour is not altered.

For more nice renders using these variants of Worley see [here](https://rmanwiki.pixar.com/display/REN26/PxrWorley) and [here](https://www.sidefx.com/docs/houdini/nodes/vop/unifiednoise.html)

worley

### DIFF
--- a/generative/worley.glsl
+++ b/generative/worley.glsl
@@ -1,5 +1,5 @@
 #include "random.glsl"
-#include "../math/distance.glsl"
+#include "../math/dist.glsl"
 
 /*
 contributors: Patricio Gonzalez Vivo
@@ -17,8 +17,8 @@ license:
 #ifndef FNC_WORLEY
 #define FNC_WORLEY
 
-#ifndef DIST_FNC
-#define DIST_FNC euclidean
+#ifndef WORLEY_DIST_FNC
+#define WORLEY_DIST_FNC distEuclidean
 #endif
 
 float worley(vec2 p){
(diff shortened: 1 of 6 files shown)


patriciogonzalezvivo:
This is fantastic! 

Neat picking on the file `math/distance.*`. So far the pattern have been `filename = function name` this one is a bit tricky  because `distance` exist as a native function name. 

Questions:

1. Are this functions reusable? 
2. Would be ok to make a `dist()` functions that have several types `distEuclidean()`, `distManhattan()`, etc.?  And by defaults `dist()` uses `distEclidean()`? a bit like: https://github.com/patriciogonzalezvivo/lygia/blob/main/color/dither.glsl,  https://github.com/patriciogonzalezvivo/lygia/blob/main/filter/edge.glsl or https://github.com/patriciogonzalezvivo/lygia/blob/main/color/tonemap.glsl
3. If is not really reusable I think makes more sense to wrap it on the worley functions like: https://github.com/patriciogonzalezvivo/lygia/blob/main/draw/colorChecker.glsl

shadielhajj:
@patriciogonzalezvivo Totally reusable! And useful in many scenarios. Your naming suggestion makes sense. I'll amend.

shadielhajj:
@patriciogonzalezvivo naming amended :-)

patriciogonzalezvivo:
Thank you!

PR #162, opened 2024-07-10
Anti-aliased triangle wave (GLSL and HLSL)

shadielhajj:
In the spirit of previous improvements to aafract, I added an anti-aliased and band-limited variant of math/mirror. Since this is a full rethink and rewrite of the function, and to emphasize the antialiased nature of the function and similarity to aafract and co., I named it aamirror.

math/mirror
Screenshot 2024-07-10 223603

math/aamirror
Screenshot 2024-07-10 223720

### DIFF
--- a/math/aamirror.glsl
+++ b/math/aamirror.glsl
@@ -1,0 +1,23 @@
+#include "nyquist.glsl"
+
+/*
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
Love it!

PR #163, opened 2024-07-11
Moar Noise, plz!

shadielhajj:
Currently, the `worley` function returns `1.0-F1` where F1 is the distance of a given point (which coordinates are passed as input param) to the closest seed point).

I modified it slightly to return `vec2(F1, F2)`, that is, the distances to the closest and second closest seed point, respectively.
This hands us a variety of noise patterns in a single function call (and very little overhead). F1 and F2 can be combined in many fun ways, including `F2-F1` which produces a lovely terrazzo-like pattern and can be thresholded to mimic scratches and wear.

I also added a WORLEY_JITTER option to modify the randomness of the Voronoi diagram, all the way down to 0, which results in a perfectly symmetrical pattern. The default is 1.0 (which is equivalent to the current behaviour).

These examples below are all based on the euclidean distance, but we could use any distance metrics from #161 for hours of noisy fun!

worley2

### DIFF
--- a/generative/worley.glsl
+++ b/generative/worley.glsl
@@ -21,38 +21,72 @@ license:
 #define WORLEY_DIST_FNC distEuclidean
 #endif
 
-float worley(vec2 p){
+vec2 worley(vec2 p){
     vec2 n = floor( p );
     vec2 f = fract( p );
 
-    float dis = 1.0;
-    for( int j= -1; j <= 1; j++ )
+    float distF1 = 1.0;
+    float distF2 = 1.0;
+    vec2 off1, pos1;
+    vec2 off2, pos2;
+    for( int j= -1; j <= 1; j++ ) {
         for( int i=-1; i <= 1; i++ ) {	
-                vec2  g = vec2(i,j);
-                vec2  o = random2( n + g );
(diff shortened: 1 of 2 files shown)


generative/worley.glsl
```diff
@@ -17,42 +18,80 @@ license:
...
+                distF2 = d;
+                off2 = g;
+                pos2 = p;
+            }
+        }
     }
 
-    return 1.0-dis;
+    return vec2(distF1, distF2);
 }
```
patriciogonzalezvivo, line 61: Worry a bit for back compatibility. Do you mind renaming this function from `vec2 worley(vec2 p)` to `vec2 worley2(vec2 p)`. Then creating a new `vec2 worley(vec2 p)` that just extracts the F1 of it?

```glsl
float worley(vec2 p) {
   return worley2(p).x;
}
```

patriciogonzalezvivo:
Very nice! Great idea. Super useful too!

generative/worley.glsl
shadielhajj: Done! 😊

patriciogonzalezvivo:
Thank you!

PR #166, opened 2024-07-15
Updated HLSL samplers to DirectX 10 syntax

shadielhajj:
Currently `SAMPLER_TYPE` is defined in HLSL as sampler2D, which is the obsolete DirectX 9 syntax from 2002.

This works fine with the old Unity Built-In render pipeline (BRP), which is itself deprecated.

However, it is not compatible with the newer URP and HDRP pipelines which are meant to supersede BRP. URP and HDRP use the Direct X 10 syntax (released in 2006).

Direct X 10 replaces the old `sampler2D` with Texture objects (eg. `Texture2D`) which also provide added benefits, like decoupling the texture object from the sampler state, querying size and max lod, etc...

Unreal also uses the DirectX 10 syntax exclusively.

This means texture-related functions in Lygia are currently unusable with Unity URP, Unity HDRP and Unreal.

I have updated the samplers to the DirectX 10 syntax in order to remedy this issue (which is also what has been done in the [Lygia Unreal Examples repo](https://github.com/franklzt/lygia_unreal_engine_examples)).

I have defined a default sample state `DEFAULT_SAMPLER_STATE defaultLinearClampSampler`, which means `SAMPLER_FNC(TEX, UV)` still takes two arguments and no other change is required in Lygia.

The new syntax is perfectly compatible with Unity BRP, however the examples need to be slightly updated (`sampler2D` becomes `Texture2D`), I'll follow up with a PR of the Unity examples as soon as this one is merged.

### DIFF
--- a/lighting/envMap.hlsl
+++ b/lighting/envMap.hlsl
@@ -1,6 +1,5 @@
-
+#include "../sampler.hlsl"
 #include "../math/powFast.hlsl"
-#include "../color/tonemap.hlsl"
 
 #include "fakeCube.hlsl"
 #include "toShininess.hlsl"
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
What about backwards compatibility? It's possible to through macros to request the version number like https://github.com/patriciogonzalezvivo/lygia/blob/main/sampler.glsl#L9

shadielhajj:
@patriciogonzalezvivo The new syntax is compatible with all Unity pipelines, so that shouldn't be an issue. I have amended using conditional macros anyway, just in case somebody prefers the old syntax for whichever reason.

patriciogonzalezvivo:
Thank you!!

PR #167, opened 2024-07-16
New feature: calculate normal from height map (GLSL and HLSL)

shadielhajj:
When displacing a grid based on a height map texture, we need to calculate the normals. This function does it for us.

Here's an example of 2048x2048 grid being displaced by a simple noise texture. Normals are calculated directly from the height maps using sample/normalFromTexture.
Screenshot 2024-07-16 224230

### DIFF
--- a/sample.glsl
+++ b/sample.glsl
@@ -29,3 +29,4 @@
 #include "sample/opticalFlow.glsl"
 #include "sample/smooth.glsl"
 #include "sample/zero.glsl"
+#include "sample/normalFromTexture.glsl"
--- a/sample.hlsl
+++ b/sample.hlsl
@@ -29,3 +29,4 @@
 #include "sample/opticalFlow.hlsl"
 #include "sample/smooth.hlsl"
 #include "sample/zero.hlsl"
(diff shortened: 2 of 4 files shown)


patriciogonzalezvivo:
Nice one! 
Just for the record. You can also use this one https://lygia.xyz/sample/bumpMap ( which calculates derivatives using this other one https://lygia.xyz/sample/derivative) it makes more samples. I like your approach.

Thank you again!

shadielhajj:
Thanks for the head up @patriciogonzalezvivo. I missed sample/bumpMap. Works nicely indeed.
HLSL version of sample/bumpMap is broken btw. I'll make a PR with a fix in a sec.

patriciogonzalezvivo:
Nit: I was wondering about the name 'sampleNormalFromTextue' wouldn't be more clear 'sampleNormalFromHeightmap'? Being a sampler implies is sampling from a texture. Just a thought. Happy to go with your gut feeling

shadielhajj:
@patriciogonzalezvivo agreed. I was also hesitating. Let's go with `normalFromHeightmap`. I'll PR.

PR #170, opened 2024-07-17
Fixed raymarching camera (HLSL)

shadielhajj:
Ok, this one gave me a hard time 😅

Raymarching camera was broken in hlsl and rotating in erratic ways when moved.
Turns out the culprit was matrix layout in `raymarch/camera.hlsl`:
glsl is row-major anh hlsl column major. The returned `float3x3` is initialised with row vectors, but `cu`, `cv` and `cw` are column vectors. Thus, we need to transpose the matrix before returning it (HLSL only).

There might def be similar errors elsewhere in the HLSL, I'll try to find a moment to track them down.

### DIFF
--- a/lighting/raymarch.hlsl
+++ b/lighting/raymarch.hlsl
@@ -28,6 +28,14 @@ license:
 #define RESOLUTION _ScreenParams.xy
 #endif
 
+#ifndef RAYMARCH_CAMERA_FOV
+#define RAYMARCH_CAMERA_FOV 3.0
+#endif
+
+#ifndef RAYMARCH_CAMERA_SCALE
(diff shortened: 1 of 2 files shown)


patriciogonzalezvivo:
oh my!! Sorry for that. Yes, there must be more row-major matrixes in HLSL

: |

Sorry

shadielhajj:
No worries @patriciogonzalezvivo, it's one of these obscure gotchas when porting GLSL to HLSL.
For future reference, this is useful: https://learn.microsoft.com/en-us/windows/uwp/gaming/glsl-to-hlsl-reference

PR #171, opened 2024-07-22
feat: add cnoise wgsl

dmnsgn:
I originally ported it in my WebGPU engine a few years back:
https://github.com/dmnsgn/dgel/blob/main/src/shaders/noise/classic.wgsl.ts

And tested it with the following example: https://dmnsgn.github.io/dgel/?id=instancing

Let me know if I need to update the contributors list or any other form of docs.


patriciogonzalezvivo:
Thank you!

PR #173, opened 2024-07-23
feat: add pnoise wgsl

dmnsgn:
Also cleanup wgsl noise use docs types (eg. use vec2f instead of vec2)


patriciogonzalezvivo:
Thank you!

PR #180, opened 2024-07-25
add: srandom, noised wgsl

sdedovic:
A few questions:

- `srandom_tile2`, `srandom_tile3` are not in the HLSL file but are in the GLSL file. Should they be ported here?
- I added myself as the contributor to each and copied the names in the GLSL versions (as I copied and formatted those)

### DIFF
--- a/generative/noised.wgsl
+++ b/generative/noised.wgsl
@@ -3,7 +3,6 @@
 /*
 contributors:
     - Inigo Quilez
-    - Stevan Dedovic
 description: Returns 2D/3D value noise in the first channel and in the rest the derivatives. For more details read this nice article http://www.iquilezles.org/www/articles/gradientnoise/gradientnoise.htm
 use: noised(<vec2|vec3> space)
 options:
@@ -11,17 +10,6 @@ options:
 examples:
     - /shaders/generative_noised.frag
 */
-
-#ifndef NOISED_RANDOM2_FNC
-#define NOISED_RANDOM2_FNC srandom22
-#endif
-
-#ifndef NOISED_RANDOM3_FNC
-#define NOISED_RANDOM3_FNC srandom33
-#endif
-
-#ifndef FNC_NOISED
-#define FNC_NOISED
 
 // return gradient noise (in x) and its derivatives (in yz)
 fn noised2 (p: vec2f) -> vec3f {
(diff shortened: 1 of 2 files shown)


sdedovic:
I was able to (manually) preprocess this file and compile it into one of my projects. I have fully tested the functions, though.

patriciogonzalezvivo:
Hi @sdedovic! I'm so glad to see you here! 
To your first question, yes. GLSL is usually a head or consider cannon. Lately @shadielhajj have been making huge contributions to the HLSL side, specially for the PBR and Raymarching functions.   
To your second question, I led that to your discretion. I usually don't add my self when is just a translation. And just leave the original authors. But sometimes that process is not a 1:1 and feel the result is a different thing that the originally and in that case I do add my self.

I have a question about macros. What are you using to compile in WGSL? Are they supported or made your own implementation. I was working on WGSL heavily last year and at that point we didn't have `#defines` so I didn't use them. As you can tell for this library I really really love them and find them useful. Is something we can relay for name collisions or assume users can count on them to customize function behavior?

Thank you so much for bringing some TLC to WGSL side of LYGIA

generative/noised.wgsl
```diff
@@ -0,0 +1,86 @@
...
+    - Stevan Dedovic
+description: Returns 2D/3D value noise in the first channel and in the rest the derivatives. For more details read this nice article http://www.iquilezles.org/www/articles/gradientnoise/gradientnoise.htm
+use: noised(<vec2|vec3> space)
+options:
+    NOISED_QUINTIC_INTERPOLATION: Quintic interpolation on/off. Default is off.
+examples:
+    - /shaders/generative_noised.frag
+*/
+
+#ifndef NOISED_RANDOM2_FNC
```
patriciogonzalezvivo, line 15: Assuming for your comment on "(manually) preprocessing" and compiling this shader I assume `#define` are still not valid WGSL. In that case let's take all this `#ifdef`, `#ifndef` and `#define` calls. Same with the mention on the commented header

sdedovic: > `#define` are still not valid WGSL
That is correct.

Do you mean I should remove the `#ifdef`, `#ifndef` and `#define`? I agree, as it is not useful.

As for removing the `#include`, I would rather leave that as a comment, e.g. `// #include "../math/mod.wgsl"` so that there is a reference to the dependency chain. Or maybe a different top-level comment to simply state, "this file depends on this other file".

Thoughts?

sdedovic:
> I have a question about macros. What are you using to compile in WGSL?

I have a WGPU (rust-lang) project and I believe it uses [Naga](https://github.com/gfx-rs/wgpu/tree/trunk/naga) internally to translate the WGPU shader to the appropriate language. My project has integration tests that I manually run on Windows (DirectX), Linux (Vulkan / OpenGL), and MacOS (Metal).

As for preprocessing, I manually copy everything into one file. I plan to write my own pre-processor and open source it. Maybe in the next month.

generative/noised.wgsl
patriciogonzalezvivo: Exactly all preprocessing macros except '#include's

sdedovic:
@patriciogonzalezvivo updated

patriciogonzalezvivo (approved):

patriciogonzalezvivo:
Awesome! thank you!

Wellcome to the license!
Screenshot 2024-08-01 at 11 48 08 AM

PR #182, opened 2024-07-25
PBR: fallback to environment sampling if Spherical Harmonics are not present.

shadielhajj:
Currently, the PBR shader expects SCENE_SH_ARRAY (the spherical harmonics array) to be defined and will fail if not (the indirect diffuse component is broken).
An elegant fallback is to sample the environment at a high LOD level.


patriciogonzalezvivo:
Smart improvement! Thank you

PR #183, opened 2024-07-26
Polar coordinates

shadielhajj:
* Fixed cart2cart.glsl 3D (misnamed angles)
* Updated cart2cart.hlsl 3D to match glsl implementation
* Added polar2cart 3D (glsl and hlsl)
* Added degree/radians conversion constants.


patriciogonzalezvivo:
<3

PR #185, opened 2024-07-29
Fixes and improvements to environment sampling (PBR)

shadielhajj:
* Added trilinear sampling (HLSL)
* Fixed GLSL environment sampling functions
* Added `ENVMAP_LOD_OFFSET` option
* Tweaked diffuse irradiance component


lighting/pbr.glsl
```diff
@@ -85,7 +85,7 @@ vec4 pbr(const in Material _mat) {
     #if defined(SCENE_SH_ARRAY)
     Fd  *= tonemap( sphericalHarmonics(M.normal) );
     #else
-    Fd *= envMap(M.normal, 0.8); // a roughness of 0.8 visually matches the factors returned by Unity's ShadeSH9
+    Fd *= envMap(M.normal, 1);
```
patriciogonzalezvivo, line 88: `1` -> `1.0`

lighting/envMap.glsl
```diff
@@ -21,16 +22,20 @@ license:
 
 #ifndef SAMPLE_CUBE_FNC
 #if __VERSION__ >= 300
-#define SAMPLE_CUBE_FNC(CUBEMAP, NORM, LOD) texture(CUBEMAP, NORM, LOD)
+#define SAMPLE_CUBE_FNC(CUBEMAP, NORM, LOD) textureLod(CUBEMAP, NORM, LOD)
 #else
-#define SAMPLE_CUBE_FNC(CUBEMAP, NORM, LOD) textureCube(CUBEMAP, NORM, LOD)
+#define SAMPLE_CUBE_FNC(CUBEMAP, NORM, LOD) textureCubeLod(CUBEMAP, NORM, LOD)
```
patriciogonzalezvivo, line 27: Noticing that GLSL 120 don't support textureCubeLod(CUBEMAP, NORM, LOD) I'm reverting this one to textureCube(CUBEMAP, NORM, LOD).

shadielhajj: @patriciogonzalezvivo These functions are not equivalent and do different things. `textureCube` is incorrect in that case.

shadielhajj: @patriciogonzalezvivo is it an option to use this extension? https://developer.mozilla.org/en-US/docs/Web/API/EXT_shader_texture_lod

patriciogonzalezvivo: Apparently requires version 300.

"This extension is only available to WebGL1 contexts. In WebGL2, the functionality of this extension is available on the WebGL2 context by default. It requires GLSL #version 300 es."

Should we add a sampleCubeLod.glsl that gracefully replaces this under 300?

shadielhajj: Sure. I can do a bit more digging later.

PR #190, opened 2024-08-01
Overhaul of the Raymarching material system.

shadielhajj:
This is part of a pretty thorough overhaul of the raymarching material framework. Client-side modifications are required, but we tried to keep them manageable.

The objectives are:
* Make the system more flexible and amenable to evolution
* Allow multiple materials per scene
* Integrate with Lygia mesh-based shading pipelines, starting with PBR

The previous material system relies on a single `vec4` which encodes colour and distance (from SDF).
* Material properties are not supported on the fragment level. eg. everything in the scene is metallic or not.
* No out-of-the-box integration with existing models

We replace this `vec4` with the existing `Material` struct.
* Every single fragment now specifies its own material properties. We can mix and match metallic/dielectric smooth/rough objects, at the fragment level.
* Tightly integrated with existing shading algorithm. Switching from the default raymarching shading model to the super-sexy Lygia PBR model is as easy as adding: `#define RAYMARCH_SHADING_FNC pbr`

Example raymarching map function (default shader):

Screenshot 2024-08-01 134834

```glsl
Material raymarchMap( in vec3 pos ) {
    Material res = materialNew();
    res.sdf = RAYMARCH_MAX_DIST;

    float check = checkBoard(pos.xz, vec2(1.0, 1.0));

    res = opUnion( res, materialNew(0.5 + vec3(check, check, check) * 0.5, 0.0, 0.5, planeSDF(pos)));

    res = opUnion( res, materialNew( vec3(1.0, 1.0, 1.0), sphereSDF(   pos-vec3( 0.0, 0.60, 0.0), 0.5 ) ) );
    res = opUnion( res, materialNew( vec3(0.0, 1.0, 1.0), boxSDF(      pos-vec3( 2.0, 0.5, 0.0), vec3(0.4, 0.4, 0.4) ) ) );
    res = opUnion( res, materialNew( vec3(0.3, 0.3, 1.0), torusSDF(    pos-vec3( 0.0, 0.5, 2.0), vec2(0.4,0.1) ) ) );
    res = opUnion( res, materialNew( vec3(0.3, 0.1, 0.3), capsuleSDF(  pos,vec3(-2.3, 0.4,-0.2), vec3(-1.6,0.75,0.2), 0.2 ) ) );
    res = opUnion( res, materialNew( vec3(0.5, 0.3, 0.4), triPrismSDF( pos-vec3(-2.0, 0.50,-2.0), vec2(0.5,0.1) ) ) );
    res = opUnion( res, materialNew( vec3(0.2, 0.2, 0.8), cylinderSDF( pos-vec3( 2.0, 0.50,-2.0), vec2(0.2,0.4) ) ) );
    res = opUnion( res, materialNew( vec3(0.7, 0.5, 0.2), coneSDF(     pos-vec3( 0.0, 0.75,-2.0), vec3(0.8,0.6,0.6) ) ) );
    res = opUnion( res, materialNew( vec3(0.4, 0.2, 0.9), hexPrismSDF( pos-vec3(-2.0, 0.60, 2.0), vec2(0.5,0.1) ) ) );
    res = opUnion( res, materialNew( vec3(0.1, 0.3, 0.6), pyramidSDF(  pos-vec3( 2.0, 0.10, 2.0), 1.0 ) ) );;

    return res;
}
``` 
Example raymarching map function (multiple PBR materials):

Screenshot 2024-08-01 145856

```glsl

#define RAYMARCH_SHADING_FNC pbr

Material raymarchMap( in vec3 pos ) {
    Material res = materialNew();
    res.sdf = RAYMARCH_MAX_DIST;

    float check = checkBoard(pos.xz, vec2(1.0, 1.0));

    res = opUnion( res, materialNew(0.5 + vec3(check, check, check) * 0.5, 0.0, 0.5, planeSDF(pos)));

    res = opUnion( res, materialNew( vec3(1.0, 1.0, 1.0), 1.0, 0.0, sphereSDF(   pos-vec3( 0.0, 0.60, 0.0), 0.5 ) ) );
    res = opUnion( res, materialNew( vec3(0.0, 1.0, 1.0), 1.0, 0.2, boxSDF(      pos-vec3( 2.0, 0.5, 0.0), vec3(0.4, 0.4, 0.4) ) ) );
    res = opUnion( res, materialNew( vec3(0.3, 0.3, 1.0), 1.0, 0.0, torusSDF(    pos-vec3( 0.0, 0.5, 2.0), vec2(0.4,0.1) ) ) );
    res = opUnion( res, materialNew( vec3(0.3, 0.1, 0.3), 1.0, 0.0, capsuleSDF(  pos,vec3(-2.3, 0.4,-0.2), vec3(-1.6,0.75,0.2), 0.2 ) ) );
    res = opUnion( res, materialNew( vec3(0.5, 0.3, 0.4), 0.0, 0.0, triPrismSDF( pos-vec3(-2.0, 0.50,-2.0), vec2(0.5,0.1) ) ) );
    res = opUnion( res, materialNew( vec3(0.2, 0.2, 0.8), 0.0, 0.0, cylinderSDF( pos-vec3( 2.0, 0.50,-2.0), vec2(0.2,0.4) ) ) );
    res = opUnion( res, materialNew( vec3(0.7, 0.5, 0.2), 1.0, 0.1, coneSDF(     pos-vec3( 0.0, 0.75,-2.0), vec3(0.8,0.6,0.6) ) ) );
    res = opUnion( res, materialNew( vec3(0.4, 0.2, 0.9), 0.0, 1.0, hexPrismSDF( pos-vec3(-2.0, 0.60, 2.0), vec2(0.5,0.1) ) ) );
    res = opUnion( res, materialNew( vec3(0.1, 0.3, 0.6), 0.0, 1.0, pyramidSDF(  pos-vec3( 2.0, 0.10, 2.0), 1.0 ) ) );;

    return res;
}
```

Notice how materials are specified using the material constructor shortcut `Material materialNew(vec3 albedo, float metallic, float roughness, float sdf)`

Future work
* Integrate with the raymarching volume renderer and allow mixing volume and solid in the same scene.

### DIFF
--- a/lighting/material.glsl
+++ b/lighting/material.glsl
@@ -26,6 +26,10 @@ license:
     - Copyright (c) 2021 Patricio Gonzalez Vivo under Patron License - https://lygia.xyz/license
 */
 
+#if !defined(MATERIAL_OPT_IN)
+#define RENDER_RAYMARCHING
+#endif
+
 #ifndef STR_MATERIAL
 #define STR_MATERIAL
 struct Material {
@@ -34,8 +38,11 @@ struct Material {
 
     vec3    position;       // world position of the surface
     vec3    normal;         // world normal of the surface
+
+#if defined(RENDER_RAYMARCHING)
     float   sdf;
     bool    valid;
+#endif
 
     #if defined(SCENE_BACK_SURFACE)
     vec3    normal_back;    // world normal of the back surface of the model
--- a/lighting/material.glsl
+++ b/lighting/material.glsl
@@ -48,7 +48,7 @@ struct Material {
     float   metallic;
     float   ambientOcclusion;   // default 1.0
 
-#if defined (SHADING_MODEL_CLEAR_COAT)
+#if defined(SHADING_MODEL_CLEAR_COAT)
     float   clearCoat;
     float   clearCoatRoughness;
     #if defined(MATERIAL_HAS_CLEAR_COAT_NORMAL)
--- a/lighting/material.glsl
+++ b/lighting/material.glsl
@@ -35,6 +35,7 @@ struct Material {
     vec3    position;       // world position of the surface
     vec3    normal;         // world normal of the surface
     float   sdf;
+    bool    valid;
 
     #if defined(SCENE_BACK_SURFACE)
     vec3    normal_back;    // world normal of the back surface of the model
--- a/lighting/material.glsl
+++ b/lighting/material.glsl
@@ -14,11 +14,10 @@
 contributors: Patricio Gonzalez Vivo
 description: Generic Material Structure
 options:
-    - SURFACE_POSITION
-    - SHADING_SHADOWS
-    - MATERIAL_HAS_CLEAR_COAT
-    - MATERIAL_CLEARCOAT_ROUGHNESS
+    - SCENE_BACK_SURFACE
+    - SHADING_MODEL_CLEAR_COAT
     - MATERIAL_HAS_CLEAR_COAT_NORMAL
+    - SHADING_MODEL_IRIDESCENCE
     - SHADING_MODEL_SUBSURFACE
     - SHADING_MODEL_CLOTH
     - SHADING_MODEL_SPECULAR_GLOSSINESS
@@ -35,6 +34,7 @@ struct Material {
 
     vec3    position;       // world position of the surface
     vec3    normal;         // world normal of the surface
+    float   sdf;
 
     #if defined(SCENE_BACK_SURFACE)
     vec3    normal_back;    // world normal of the back surface of the model
@@ -46,15 +46,14 @@ struct Material {
     float   roughness;
     float   metallic;
     float   ambientOcclusion;   // default 1.0
-    // float   shadow;             // default 1.0
 
-// #if defined(MATERIAL_HAS_CLEAR_COAT)
+#if defined (SHADING_MODEL_CLEAR_COAT)
     float   clearCoat;
     float   clearCoatRoughness;
     #if defined(MATERIAL_HAS_CLEAR_COAT_NORMAL)
     vec3    clearCoatNormal;    // default vec3(0.0, 0.0, 1.0);
     #endif
-// #endif
+#endif
 
 #if defined(SHADING_MODEL_IRIDESCENCE)
     float   thickness; // default to 300.0
--- a/lighting/material/new.glsl
+++ b/lighting/material/new.glsl
@@ -117,14 +117,14 @@ void materialNew(out Material _mat) {
 #endif
 
 #if defined(SHADING_MODEL_SPECULAR_GLOSSINESS)
-    vec3    specularColor;
-    float   glossiness;
+    _mat.specularColor = vec3(0.04, 0.04, 0.04);
+    _mat.glossiness = 0.5;
 #endif
 
     // Cache
-    vec3    V;
-    vec3    R;
-    float   NoV;
+    _mat.V = vec3(0.0);
+    _mat.R = vec3(0.0);
+    _mat.NoV = 0.0;
 }
 
 Material materialNew() {
--- a/lighting/material/new.glsl
+++ b/lighting/material/new.glsl
@@ -44,6 +44,7 @@ void materialNew(out Material _mat) {
     _mat.position           = (SURFACE_POSITION).xyz;
     _mat.normal             = materialNormal();
     _mat.sdf                = 0.0;
+    _mat.valid              = true;
 
     #if defined(SCENE_BACK_SURFACE) && defined(RESOLUTION)
         vec4 back_surface       = SAMPLER_FNC(SCENE_BACK_SURFACE, gl_FragCoord.xy / RESOLUTION);
@@ -129,6 +130,13 @@ void materialNew(out Material _mat) {
 Material materialNew() {
     Material mat;
     materialNew(mat);
+    return mat;
+}
+
+Material materialNew(float3 albedo, float sdf) {
+    Material mat = materialNew();
+    mat.albedo.rgb = albedo;
+    mat.sdf = sdf;
     return mat;
 }
 
--- a/lighting/material/new.glsl
+++ b/lighting/material/new.glsl
@@ -133,14 +133,14 @@ Material materialNew() {
     return mat;
 }
 
-Material materialNew(float3 albedo, float sdf) {
+Material materialNew(vec3 albedo, float sdf) {
     Material mat = materialNew();
     mat.albedo.rgb = albedo;
     mat.sdf = sdf;
     return mat;
 }
 
-Material materialNew(float3 albedo, float metallic, float roughness, float sdf) {
+Material materialNew(vec3 albedo, float metallic, float roughness, float sdf) {
     Material mat = materialNew();
     mat.albedo.rgb = albedo;
     mat.metallic = metallic;
--- a/lighting/material/new.glsl
+++ b/lighting/material/new.glsl
@@ -43,8 +43,11 @@ void materialNew(out Material _mat) {
     // Surface data
     _mat.position           = (SURFACE_POSITION).xyz;
     _mat.normal             = materialNormal();
+
+#if defined(RENDER_RAYMARCHING)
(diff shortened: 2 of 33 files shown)


patriciogonzalezvivo:
WOW. LOVE the goals. I'm so here for it. Thank you for executing on this vision. LOVE IT. pluss looks SO good
I will run this first on my OpenGL ES 2.0 pipeline

lighting/material.glsl
```diff
@@ -35,6 +34,8 @@ struct Material {
 
     vec3    position;       // world position of the surface
     vec3    normal;         // world normal of the surface
+    float   sdf;
```
patriciogonzalezvivo, line 37: Maybe adding `#if defined(RENDER_RAYMARCHING)` and to `raymarching.*` add this tag like with `pbrClearCoat.*` ?

sdf/opIntersection.glsl
```diff
@@ -11,6 +11,15 @@ use: <float> opIntersection( in <float> d1, in <float> d2 [, <float> smooth_fact
 
 float opIntersection( float d1, float d2 ) { return max(d1,d2); }
 
+Material opIntersection(Material d1, Material d2)
```
patriciogonzalezvivo, line 14: Maybe add:

#if defined(STR_MATERIAL)

in case people use this function without material definition?
or
add #include "material.glsl" at the top

sdf/opUnion.glsl
```diff
@@ -10,7 +10,14 @@ use: <float> opUnion( in <float|vec4> d1, in <float|vec4> d2 [, <float> smooth_f
 #define FNC_OPUNION
 
 float opUnion( float d1, float d2 ) { return min(d1, d2); }
-vec4  opUnion( vec4 d1, vec4 d2 ) { return (d1.a < d2.a) ? d1 : d2; }
+
+Material opUnion( Material d1, Material d2 ) {
```
patriciogonzalezvivo, line 14: Maybe add:

#if defined(STR_MATERIAL)

in case people use this function without material definition?
or
add #include "material.glsl" at the top

lighting/material/new.glsl
```diff
@@ -111,16 +108,23 @@ void materialNew(out Material _mat) {
...
     #endif
 
 #endif
 
-    // Cloth Model
 #if defined(SHADING_MODEL_CLOTH)
     _mat.sheenColor         = sqrt(_mat.albedo.rgb);
 #endif
+
+#if defined(SHADING_MODEL_SPECULAR_GLOSSINESS)
```
patriciogonzalezvivo, line 119: Uninitialized variables produce errors on some drivers.

Add:

```
#if defined(SHADING_MODEL_SPECULAR_GLOSSINESS)
    _mat.specularColor = vec3(0.04, 0.04, 0.04);
    _mat.glossiness = 0.5;
#endif

    // Cache
    _mat.V = vec3(0.0);
    _mat.R = vec3(0.0);
    _mat.NoV = 0.0;
```

lighting/pbr.glsl
```diff
@@ -52,7 +52,9 @@ vec4 pbr(const in Material _mat) {
 
     // Cached
     Material M  = _mat;
-    M.V         = normalize(CAMERA_POSITION - M.position);  // View
+    if (length(M.V) == 0.0) {
```
patriciogonzalezvivo, line 55: wonder if there is a more efficient way of doing this check. Like
```
if (M.V.x != 0.0 && M.V.y != 0.0 & M.V.z != 0.0) {
```

lighting/material/new.glsl
```diff
@@ -43,6 +43,8 @@ void materialNew(out Material _mat) {
     // Surface data
     _mat.position           = (SURFACE_POSITION).xyz;
     _mat.normal             = materialNormal();
+    _mat.sdf                = 0.0;
```
patriciogonzalezvivo, line 46: Maybe:

```
_mat.sdf = RAYMARCH_MAX_DIST;
```

So it doesn't need to be define on `raymarchMap()`.

This is bringing a pattern that rendering method needs to be define first. Same for `pbrClearCoat()` We should need to add this on some documentation

patriciogonzalezvivo, line 46: Add `#if defined(RENDER_RAYMARCHING)` or `#if defined(SHADING_MODEL_SDF)` or similar. What ever seems better for you

patriciogonzalezvivo (approved):
Ok, I give it a first pass. Made commits for typos and comments for suggestions or things to think about. 
Hope this style is ok for you. Let me know.

lighting/material.glsl
shadielhajj: It's a good idea, but it makes the include system a bit more brittle. If `material.glsl` is included before `raymarch.glsl`(on the client side), then `RENDER_RAYMARCHING`will not get a chance to be declared, and the taymarching system will break. Let me give it a think and get back to you.

shadielhajj: @patriciogonzalezvivo due to circular dependencies, this is turning out to be more complicated than hoped.

Take this simple example:
```
#include "lygia/lighting/pbr.glsl"
#include "lygia/lighting/raymarch.glsl"
```
`pbr.glsl` includes `material.glsl` which means `Material` will get initialised before `raymarch.glsl` gets a chance to define `RENDER_RAYMARCHING`.

The reverse option also fails
```
#include "lygia/lighting/raymarch.glsl"
#include "lygia/lighting/pbr.glsl"
```
since `raymarch.glsl` relies on the `pbr` shading function.

if we  really want to use the `#if defined(RENDER_RAYMARCHING)` pattern, the only option I can think of is to require the client to manually define `RENDER_RAYMARCHING` in their code when needed:
```
#define RENDER_RAYMARCHING
#include "lygia/lighting/raymarch.glsl"
#include "lygia/lighting/pbr.glsl"
```

I'm OK with it. Let me know what you think.

patriciogonzalezvivo: Good points. So it seams is about managing memory foot print vs easiness. Let's favor easiness and leave room for memory optimization.

What about adding this at the top:

```glsl
#if !defined(MATERIAL_OPT_IN)
#define RENDER_RAYMARCHING
#define SHADING_MODEL_CLOTH
#define SHADING_MODEL_CLEAR_COAT
#define SHADING_MODEL_CLOTH
#define SHADING_MODEL_SPECULAR_GLOSSINESS
...
#endif
``` 
Where by default is permissive and instantiate all the variables on the struct. And then for pro-users that want to be more efficient they need to use `#define MATERIAL_OPT_IN` + the defines they want

shadielhajj: @patriciogonzalezvivo Sounds like a good compromise! Just pushed to the branch.
BTW it's only needed for `RENDER_RAYMARCHING`. `SHADING_MODEL_*` all live in material sub-system and cause no circular dependency issues.

lighting/material/new.glsl
shadielhajj: @patriciogonzalezvivo  That's also a good idea, but subject to the same circular dependency issues. Shall we park it for now?

shadielhajj:
@patriciogonzalezvivo Are we ready to merge this one, or is there something else I overlooked or we need to look at?

patriciogonzalezvivo:
Let's merge it! Go a head. Or I can early tomorrow

shadielhajj:
@patriciogonzalezvivo I don't think I have permission to merge. You'll have to do it. No rush though :-)

patriciogonzalezvivo:
@shadielhajj I notice this regression on my side. Do you notice something similar?

image
https://github.com/patriciogonzalezvivo/lygia_examples/blob/main/lighting_pbr.frag

patriciogonzalezvivo:
Found the culprit! Was
```glsl
if (M.V.x == 0.0 && M.V.y == 0.0 && M.V.z == 0.0) {
```
instead of
```glsl
if (M.V.x != 0.0 && M.V.y != 0.0 && M.V.z != 0.0) {
```
: P

patriciogonzalezvivo:
Pushing a fix

shadielhajj:
Thanks, @patriciogonzalezvivo! My bad.

PR #197, opened 2024-08-13
Feature/metal - some sampler + filter additions 

vade:
Finally had a moment to work on this.

* had to add a `SAMPLER` which metal uses, which defaults to `sampler( min_filter::linear, mag_filter::linear )`
* set the default texture2d precision to float to keep things in parity with the rest of Lygia. This means that metal shaders need to define their texture as float, not half which is maybe more standard. You can override this by redefining `SAMPLER_TYPE`

Implemented box blur and gaussian blur just to test the waters.

### DIFF
--- a/filter/boxBlur.msl
+++ b/filter/boxBlur.msl
@@ -1,0 +1,78 @@
+#include "../sampler.msl"
+
+/*
+contributors: Patricio Gonzalez Vivo
+description: |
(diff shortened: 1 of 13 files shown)


vade:
Lygia gaussian blur running in Metal in custom node app

image

patriciogonzalezvivo:
Fantastic! That node graph of yours is pure 🔥🔥🔥

PR #198, opened 2024-08-13
Improved Raymarching view sub-system.

shadielhajj:
The previous raymarching view subsystem forces a look-at camera (a camera which calculates it's orientation to look at a specified target point). This is useful in some cases, but very awkward in others (such as fly-overs, landscapes, etc..).

The new system allows us to directly specify a view matrix. This can be used to:
1. Use a view matrix calculated by the host engine (eg. a game engine).
2. Calculate our own view matrix from a variety of functions.

Currently implemented view matrix functions:
1. lookAtViewMatrix (the previous behaviour)
2. viewMatrix (a typical camera specified using camera position and rotation)

This leaves the system open to extension with more camera matrix functions such as orbital cameras, etc...

For convenience and backward compatibility, we include a helper functions which replicates the previous behaviour.
```glsl
vec4 raymarch(vec3 cameraPosition, vec3 cameraLookAt, vec2 st)
{
    float depth = 0.0;
    vec3 worldPos = vec3(0.0, 0.0, 0.0);
    vec3 worldNormal = vec3(0.0, 0.0, 0.0);
    mat4 viewMatrix = lookAtViewMatrix(cameraPosition, cameraLookAt);
    return raymarch(viewMatrix, st, depth, worldPos, worldNormal);
}
``` 

To use a different camera function, such as a basic rotatable camera
```glsl
mat4 viewMatrix = viewMatrix(CameraPosition, CameraRotation*DEG2RAD);
vec4 renderOut = raymarch(viewMatrix, st, eyeDepth, worldPosition, worldNormals);
```


patriciogonzalezvivo:
This is great! Love the direction, less ray marching-specifics, more useful and reusable functions we can use in other cases.

NIT: do you mind dropping the `...Matrix` or `lookAtViewMatrix.*` and `viewMatrix.*`. Most of the functions in `space/` return matrices. For now we are not adding that into the names. But is a good point to have in mind. Specially if we stumble with some name collision.

shadielhajj:
> This is great! Beautiful work! Love the direction, less ray marching-specifics, more useful and reusable functions we can use in other cases.

Exactly!! :-D
 
> NIT: do you mind dropping the `...Matrix` or `lookAtViewMatrix.*` and `viewMatrix.*`. Most of the functions in `space/` return matrices. For now we are not adding that into the names. But is a good point to have in mind. Specially if we stumble with some name collision.

Sure. Just renamed them to `lookAtView` and `eulerView` (also considered `dofView`, for Degrees of Freedom, but felt it's a bit obscure).

PR #203, opened 2024-08-15
Fixed viewMatrix issue (GLSL) and Raymarching bug (HLSL)

shadielhajj:
* Fixed viewMatrix column order issue in GLSL. This was causing the cameraForward to be incorrectly calculated and breaking eyeDepth.
* Fixed HLSL bug caused by https://github.com/patriciogonzalezvivo/lygia/commit/43fc57f9b645182dcd155052e90fb345ec76f94e. Gentle reminder to always make global `const` variables `static const` in HLSL. Otherwise they will be optimised away by the compiler and cause the shader to silently fail.

### DIFF
--- a/lighting/raymarch.glsl
+++ b/lighting/raymarch.glsl
@@ -54,8 +54,8 @@ vec4 raymarch(  mat4 viewMatrix, vec2 st
     ) {
                     
     float fov = 1.0 / tan(RAYMARCH_CAMERA_FOV * DEG2RAD * 0.5);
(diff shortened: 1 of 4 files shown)


patriciogonzalezvivo:
Thank you and sorry again for the inconvenience

shadielhajj:
No worries @patriciogonzalezvivo! We got to the bottom of the view matrix issue, which was coming back to bite us at some point 😊

PR #206, opened 2024-08-16
Overhaul of volume Raymarching

shadielhajj:
The current volume raymarching sub-system is quite limited: It is not physically accurate, it's not very flexible (it doesn't support specifying volume density and works with directional light only), and most importantly, it cannot co-exist with other raymarching renderers (such as PBR). The entire scene is rendered either as a solid or a volume, not both.

The proposed rewrite:
1. First steps towards a physically accurate system
2. Per-fragment volume density
3. Supports directional lights and point lights.
4. Support for opaque objects shadowing (see below)
5. Integrated with other raymarching renderers, such as PBR

Here's a scene rendered with PBR materials and and an fbm volume 
Screenshot 2024-08-20 154410
fbm3

The code for the volume
```glsl

#define RAYMARCH_VOLUME

Medium raymarchVolumeMap( in vec3 pos ) {
    float sdf = fbm(pos);
    vec3 scattering = vec3(1.0, 1.0, 1.0);
    vec3 absorption = vec3(0.0, 0.0, 0.0);

    return mediumNew(scattering, absorption, sdf);
}
``` 

Note how the volume renderer is enabled with `RAYMARCH_VOLUME`. The map function is implemented in `raymarchVolumeMap` (by default) which must return a `Medium`. Easy peasy.

Note also that the volume is not simply overlay on the scene, but properly integrated in space. In the example below, the volume sphere correctly sits in between the cone and the doughnut.
Screenshot 2024-08-19 232036

Another cool feature is volumetric shadowing by opaque objects. Very handy to create atmospheric effects. Enable it with `RAYMARCH_VOLUMETRIC_SHADOWS`
shadowing

Bonus Track:
Rework of the AOV interface in `raymarch.*`. Just pass the desired output parameters. If full AOV output (Material) is desired, `#define RAYMARCH_AOV` should be explicitly specified. This is for optimization purposes.

```glsl
#define RAYMARCH_AOV
vec4 raymarch(mat4 viewMatrix, vec2 st, out float eyeDepth, out Material mat);

vec4 raymarch(mat4 viewMatrix, vec2 st, out float eyeDepth);

vec4 raymarch(mat4 viewMatrix, vec2 st);
``` 

Including the legacy view (lookAt) signatures
```glsl
#define RAYMARCH_AOV
vec4 raymarch(vec3 cameraPosition, vec3 cameraLookAt, vec2 st, out float eyeDepth, out Material mat);

vec4 raymarch(vec3 cameraPosition, vec3 cameraLookAt, vec2 st, out float eyeDepth);

vec4 raymarch(vec3 cameraPosition, vec3 cameraLookAt, vec2 st);
```


patriciogonzalezvivo:
Omg! this is so impressive! I'm really excited for it

Tiny suggestion:
* renaming `VolumeMaterial` class be `Volume`
* adding a separate folder call `volume/` where `new.*lsl` lives together with other volume-related functions instead of using`VolumeNew.*lsl`

patriciogonzalezvivo:
Those volumetric shadows are 🔥!!!!

shadielhajj:
Thanks @patriciogonzalezvivo ! This whole thing ended up being significantly more work than expected 😅 but I think it was worth it.
I might tweak the algo in the future, but I think it's in a good place for now, and ready to merge once you've reviewed it.

patriciogonzalezvivo:
This is fantastic work @shadielhajj! Do you mind providing a use example?

space/lookAtView.hlsl
```diff
@@ -2,6 +2,8 @@
 contributors:  Shadi El Hajj
 description: Create a look-at view matrix
 use: <float4x4> lookAtView(in <float3> position, in <float3> target, in <float3> up)
+license: MIT License (MIT) Copyright (c) 2024 Shadi EL Hajj
+license: MIT License (MIT) Copyright (c) 2024 Shadi EL Hajj
```
patriciogonzalezvivo, line 6: repeated line

lighting/common/attenuation.glsl
```diff
@@ -0,0 +1,34 @@
...
+#endif
+
+#ifndef LIGHT_ATTENUATION_EXPONENT
+#define LIGHT_ATTENUATION_EXPONENT 2
+#endif
+
+#ifndef FNC_LIGHT_ATTENUATION
+#define FNC_LIGHT_ATTENUATION
+
+float attenuation(float dist) {
```
patriciogonzalezvivo, line 26: I think this two files (`attenuation.glsl` and `attenuation.hlsl`) should be moved from `lighting/common/` to `lighting/light/`. Following the logic of https://github.com/patriciogonzalezvivo/lygia/blob/main/lighting/light/falloff.glsl

lighting/raymarch.glsl
```diff
@@ -33,201 +33,107 @@ license:
...
-            tmp.specularColor += mat.specularColor;
-            tmp.glossiness += mat.glossiness;
-            #endif
-
-            // I don't thing nobody needs this
-            // tmp.V += mat.V;
-            // tmp.R += mat.R;
-            // tmp.NoV += mat.NoV;
+        #if RAYMARCH_AOV
+            add(matAcc, mat, matAcc);
```
patriciogonzalezvivo, line 82: nice and elegant

lighting/common/attenuation.glsl
shadielhajj: Fixed!

shadielhajj:
> This is fantastic work @shadielhajj! Do you mind providing a use example?

Thanks, @patriciogonzalezvivo!

Here's how to add a spherical fog volume to the raymarching example scene and enable volumetric shadows
```glsl
#define RAYMARCH_VOLUME
#define RAYMARCH_VOLUMETRIC_SHADOWS
Medium raymarchVolumeMap( in vec3 pos ) {
    vec3 scattering = 0.05*vec3(1.0, 1.0, 1.0);
    vec3 absorption = vec3(0.0);
    return mediumNew(scattering, absorption, sphereSDF(pos-vec3( 0.0, 0.50, 0.0), 3.0 ));
}
``` 

And here's how to add noisy volumetric fog
```glsl
#define RAYMARCH_VOLUME
Medium raymarchVolumeMap( in vec3 pos ) {
    float sdf = fbm(pos);
    vec3 scattering = vec3(1.0, 1.0, 1.0);
    vec3 absorption = vec3(0.0, 0.0, 0.0);
    return mediumNew(scattering, absorption, sdf);
}
``` 

Volumetrics are quite heavy computationally, so you might want to tweak the quality params:
```glsl
#define RAYMARCH_VOLUME_SAMPLES 64
#define RAYMARCH_VOLUME_SAMPLES_LIGHT 32
#define FBM_OCTAVES 3
```

lighting/light/attenuation.glsl
```diff
@@ -0,0 +1,34 @@
+/*
+contributors: Shadi El Hajj
+description: Light attenuation equation
+license: MIT License (MIT) Copyright (c) 2024 Shadi EL Hajj
+*/
+
+#ifndef LIGHT_ATTENUATION_CONSTANT
+#define LIGHT_ATTENUATION_CONSTANT 0
```
patriciogonzalezvivo, line 8: 0.0

lighting/light/attenuation.glsl
```diff
@@ -0,0 +1,34 @@
...
+#endif
+
+#ifndef LIGHT_ATTENUATION_LINEAR
+#define LIGHT_ATTENUATION_LINEAR 0
```
patriciogonzalezvivo, line 12: 0.0

lighting/light/attenuation.glsl
```diff
@@ -0,0 +1,34 @@
...
+#endif
+
+#ifndef LIGHT_ATTENUATION_EXPONENTIAL
+#define LIGHT_ATTENUATION_EXPONENTIAL 1
```
patriciogonzalezvivo, line 16: 1.0

lighting/light/attenuation.glsl
```diff
@@ -0,0 +1,34 @@
...
+#endif
+
+#ifndef LIGHT_ATTENUATION_EXPONENT
+#define LIGHT_ATTENUATION_EXPONENT 2
```
patriciogonzalezvivo, line 20: 2.0

patriciogonzalezvivo (changes requested):

lighting/light/attenuation.glsl
shadielhajj: I can't believe I keep forgetting these!! 😅 sorry!

shadielhajj: Done!

lighting/raymarch/volume.glsl
```diff
@@ -1,103 +1,137 @@
...
+vec3 raymarchVolume(vec3 rayOrigin, vec3 rayDirection, vec2 st, float minDist, vec3 background) {
+    vec3 scatteredLuminance = vec3(0.0, 0.0, 0.0);
+    vec3 transmittance = vec3(1.0, 1.0, 1.0);
+    float stepSize = RAYMARCH_MAX_DIST/float(RAYMARCH_VOLUME_SAMPLES);
+
+    float t = RAYMARCH_MIN_DIST;
+
+    for(int i = 0; i < RAYMARCH_VOLUME_SAMPLES; i++) {        
+        vec3 position = rayOrigin + rayDirection * t;
+        Medium res = RAYMARCH_VOLUME_MAP_FNC(position);
```
patriciogonzalezvivo, line 91:
```
#if defined(RAYMARCH_VOLUME)
Medium res = RAYMARCH_VOLUME_MAP_FNC(position);
float density = -res.sdf;
vec3 scattering = res.scattering;
vec3 extinction = res.absorption + scattering;
#else 
float density = 0.0;
vec3 scattering = vec3(0.0);
vec3 extinction = vec3(0.0);
#endif
```

lighting/raymarch/volume.glsl
```diff
@@ -1,103 +1,137 @@
...
+                vec3 L = LIGHT_COLOR * LIGHT_INTENSITY;
+            #else // no lighting
+                const float attenuationL = 1.0;
+                const vec3 shadow = 1.0;
+                const vec3 L = vec3(1.0, 1.0, 1.0);
+            #endif
+
+            #if defined RAYMARCH_ENERGY_CONSERVING
+                // energy-conserving scattering integration
+                vec3 S = L * attenuationL * shadow * density * res.scattering;
```
patriciogonzalezvivo, line 118:
```
vec3 S = L * attenuationL * shadow * density * scattering;
```

lighting/raymarch/volume.glsl
```diff
@@ -1,103 +1,137 @@
...
+                vec3 sampleExtinction = max(vec3(EPSILON, EPSILON, EPSILON), density * extinction);
+                vec3 Sint = (S - S * exp(-sampleExtinction * stepSize)) / sampleExtinction;
+                scatteredLuminance += transmittance * Sint;
+                transmittance *= exp(-sampleExtinction * stepSize);
+            #else
+                // usual scattering integration. Not energy-conserving.
+                scatteredLuminance += attenuationL * shadow * transmittance * density * res.scattering * stepSize * L;
```
patriciogonzalezvivo, line 125:
```
scatteredLuminance += attenuationL * shadow * transmittance * density * scattering * stepSize * L;
```

lighting/raymarch/volume.glsl
```diff
@@ -1,103 +1,137 @@
...
+    for (int j = 0; j < RAYMARCH_VOLUME_SAMPLES_LIGHT; j++) {                
+        vec3 positionL = position + rayDirectionL * tL;
+        #if defined(RAYMARCH_VOLUMETRIC_SHADOWS)
+            Material mat = RAYMARCH_MAP_FNC(positionL);
+            if (mat.sdf <= 0.0) {
+                return vec3(0.0, 0.0, 0.0);
             }
-            col += vec4(LIGHT_COLOR * 80. * tmp * T * Tl, 1.0);
+        #endif
+        Medium resL = RAYMARCH_VOLUME_MAP_FNC(positionL);
```
patriciogonzalezvivo, line 70:
```
#if defined(RAYMARCH_VOLUME)
        Medium resL = RAYMARCH_VOLUME_MAP_FNC(positionL);
        float densityL = -resL.sdf;
        vec3 extinctionL = resL.absorption + resL.scattering;
        transmittanceL *= exp(-densityL * extinctionL * stepSizeL);
#endif
```

lighting/raymarch/map.glsl
```diff
@@ -12,9 +13,14 @@ examples:
...
 
+#ifndef RAYMARCH_VOLUME_MAP_FNC
+#define RAYMARCH_VOLUME_MAP_FNC raymarchVolumeMap
+#endif
+
 #ifndef FNC_RAYMARCH_MAP
 #define FNC_RAYMARCH_MAP
 
 Material RAYMARCH_MAP_FNC( in vec3 pos );
+Medium RAYMARCH_VOLUME_MAP_FNC( in vec3 pos );
```
patriciogonzalezvivo, line 24:
```
#if defined(RAYMARCH_VOLUME
Medium RAYMARCH_VOLUME_MAP_FNC( in vec3 pos );
#endif
```

patriciogonzalezvivo:
Sorry for all the single comments! 
I'm adding some flags to keep parity with all raymarch examples. I'm assuming for your examples that Volumetric is only desired when defining `#define RAYMARCH_VOLUME`

Hope you don't mind the additions

lighting/raymarch/volume.glsl
shadielhajj: @patriciogonzalezvivo Not sure I see what's the issue here. `res.scattering`looks correct to me. `scaterring`is not defined.

lighting/raymarch/volume.glsl
shadielhajj: @patriciogonzalezvivo same as above. `res.scattering` looks correct to me. `scaterring` is not defined.

shadielhajj: Ah ok now I see.

shadielhajj:
@patriciogonzalezvivo.  Sorry didn't mean to delete your comment, was actually trying to delete mine.

this one makes total sense
```glsl
#if defined(RAYMARCH_VOLUME)
Medium RAYMARCH_VOLUME_MAP_FNC( in vec3 pos );
#endif
``` 

Regarding this one 
```glsl
#if defined(RAYMARCH_VOLUME)
Medium res = RAYMARCH_VOLUME_MAP_FNC(position);
float density = -res.sdf;
vec3 scattering = res.scattering;
vec3 extinction = res.absorption + scattering;
#else 
float density = 0.0;
vec3 scattering = vec3(0.0);
vec3 extinction = vec3(0.0);
#endif
``` 

I don't really like the idea of breaking the algo because a `#define` is not present, apart from the code clutter. How about something like this?

```glsl
#if defined(RAYMARCH_VOLUME)
#include "raymarch/volume.glsl"
#endif
```

patriciogonzalezvivo:
The issue with that approach is that the includes inside `raymarch/volume.glsl` could be flag "trapped" inside the define flag. 
Most of the #include parsers for GLSL, or at least all of those I have implemented, tried to do their best to prevent duplications so they get checked against a map using their absolute path.

So all this files could be consider as "already added" whine actually they are not.

```
#include "map.glsl"
#include "../light/attenuation.glsl"
#include "../../generative/random.glsl"
#include "../../math/const.glsl"
#include "../medium/new.glsl"
```

To solve that all #include parsers for python, JS and C++ need to resolve all #defines first or directly not try to add it.

What about doing this?

```glsl
#if !defined(FNC_RAYMARCH_VOLUMERENDER) && defined(RAYMARCH_VOLUME)
#define FNC_RAYMARCH_VOLUMERENDER

...
```

lighting/raymarch/volume.glsl
```diff
@@ -1,103 +1,137 @@
...
+#define RAYMARCH_VOLUME_MAP_FNC raymarchVolumeMap
 #endif
 
-#ifndef RAYMARCH_MAP_FNC
-#define RAYMARCH_MAP_FNC(POS) raymarchMap(POS)
+#ifndef RAYMARCH_VOLUME_DITHER
+#define RAYMARCH_VOLUME_DITHER 0.1
 #endif
 
 #ifndef FNC_RAYMARCH_VOLUMERENDER
```
patriciogonzalezvivo, line 55: We could change the name of this one 

```
#if !defined(FNC_RAYMARCH_VOLUME) && defined(RAYMARCH_VOLUME)
#define FNC_RAYMARCH_VOLUME
```

lighting/raymarch/volume.glsl
```diff
@@ -1,103 +1,137 @@
...
-            //Light scattering
-            #ifdef LIGHT_POSITION
-            float Tl = 1.0;
-            for (int j = 0; j < nbSampleLight; j++) {
-                float densityLight = RAYMARCH_MAP_FNC( pos + sun_direction * float(j) * tstepl ).sdf;
-                if (densityLight>0.)
-                    Tl *= 1. - densityLight * absorption/fSamples;
-                if (Tl <= 0.01)
-                    break;
+vec3 shadowTransmittance(vec3 position, vec3 rayDirectionL, float stepSizeL) {
```
patriciogonzalezvivo, line 58: SUPER NIT: just flagging that this break the design rule of filename == function name. But we break and bend that rule all the time. I think is ok. Just thinking to planting the question... on how we can keep that rule more often that not... and how to make consistent the tacit rule about function names that incorporate the previous folder

lighting/raymarch/volume.glsl
shadielhajj: @patriciogonzalezvivo  Even better!! I'll amend.

lighting/raymarch/volume.glsl
shadielhajj: @patriciogonzalezvivo interesting question. I agree we need rules for see.
The way I think of this is private vs. public functions:
* If the function is meant to be public (re-usable elsewhere), then we should def stick to `filename == function`.
* If the function is "private", as in this case (meant to be used as a helper function only called from functions inside the same file) then filename == function` does not apply.
* If you think the above rules make sense, then should be flag private function with a suffix? eg. `_shadowTransmittance`. Just a mere suggestion. I also think it can stay as it is, and the fact it's private can be deduced from `filename != function`

lighting/raymarch/volume.glsl
shadielhajj: Done!

lighting/raymarch/volume.glsl
patriciogonzalezvivo: Great point. I have been amending that principle often incorporating the "legal name" before the private function. In this case
```
raymarchVolumeShadowTransmittance()
```

The idea behind is that collisions become statistically harder, while the affordance of the function name less important because of their private nature

patriciogonzalezvivo: again.. all this is very NIT picking

shadielhajj: `raymarchVolumeShadowTransmittance()` makes total sense. Let's go with that.
Amended and pushed.

patriciogonzalezvivo (approved):

patriciogonzalezvivo:
Brillant PR @shadielhajj!

PR #212, opened 2024-08-23
Minor param refactoring in Raymarch

shadielhajj:
On a quest to remove all magic numbers from the raymarching code and expose them as meaningful parameters whenever possible:
* Exposed offset param in `normal`
* Exposed all magic number in `ao` as params
* Refactored `softShadow`params.

### DIFF
--- a/lighting/raymarch/ao.glsl
+++ b/lighting/raymarch/ao.glsl
@@ -35,8 +35,9 @@ examples:
 float raymarchAO(in vec3 pos, in vec3 nor) {
     float occ = 0.0;
     float sca = 1.0;
+    const float samplesFactor = 1.0 / float(RAYMARCH_AO_SAMPLES-1);
     for (int i = 0; i < RAYMARCH_AO_SAMPLES; i++) {
-        float h = RAYMARCH_AO_MIN_DIST + RAYMARCH_AO_MAX_DIST * float(i) / (RAYMARCH_AO_SAMPLES-1);
+        float h = RAYMARCH_AO_MIN_DIST + RAYMARCH_AO_MAX_DIST * float(i) * samplesFactor;
         float d = RAYMARCH_MAP_FNC(pos + h * nor).sdf;
         occ += (h - d) * sca;
         sca *= RAYMARCH_AO_FALLOFF;
--- a/lighting/raymarch/ao.glsl
+++ b/lighting/raymarch/ao.glsl
@@ -9,8 +9,24 @@ examples:
     - /shaders/lighting_raymarching.frag
 */
 
-#ifndef RAYMARCH_SAMPLES_AO
-#define RAYMARCH_SAMPLES_AO 5
+#ifndef RAYMARCH_AO_SAMPLES
+#define RAYMARCH_AO_SAMPLES 5
+#endif
+
+#ifndef RAYMARCH_AO_INTENSITY
+#define RAYMARCH_AO_INTENSITY 1.0
+#endif
+
+#ifndef RAYMARCH_AO_MIN_DIST
+#define RAYMARCH_AO_MIN_DIST 0.001
+#endif
+
+#ifndef RAYMARCH_AO_MAX_DIST
+#define RAYMARCH_AO_MAX_DIST 0.2
+#endif
+
+#ifndef RAYMARCH_AO_FALLOFF
+#define RAYMARCH_AO_FALLOFF 0.95
 #endif
 
 #ifndef FNC_RAYMARCH_AO
@@ -19,13 +35,13 @@ examples:
 float raymarchAO(in vec3 pos, in vec3 nor) {
     float occ = 0.0;
     float sca = 1.0;
-    for (int i = 0; i < RAYMARCH_SAMPLES_AO; i++) {
-        float h = 0.001 + 0.15 * float(i) / 4.0;
+    for (int i = 0; i < RAYMARCH_AO_SAMPLES; i++) {
+        float h = RAYMARCH_AO_MIN_DIST + RAYMARCH_AO_MAX_DIST * float(i) / (RAYMARCH_AO_SAMPLES-1);
         float d = RAYMARCH_MAP_FNC(pos + h * nor).sdf;
         occ += (h - d) * sca;
-        sca *= 0.95;
+        sca *= RAYMARCH_AO_FALLOFF;
     }
-    return clamp(1.0 - 1.5 * occ, 0.0, 1.0);
+    return saturate(1.0 - RAYMARCH_AO_INTENSITY * occ);
 }
 
 #endif
--- a/lighting/raymarch/softShadow.glsl
+++ b/lighting/raymarch/softShadow.glsl
@@ -13,8 +13,12 @@ examples:
     - /shaders/lighting_raymarching.frag
 */
 
-#ifndef RAYMARCHSOFTSHADOW_ITERATIONS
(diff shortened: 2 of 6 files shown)


lighting/raymarch/ao.glsl
```diff
@@ -19,13 +35,13 @@ examples:
 float raymarchAO(in vec3 pos, in vec3 nor) {
     float occ = 0.0;
     float sca = 1.0;
-    for (int i = 0; i < RAYMARCH_SAMPLES_AO; i++) {
-        float h = 0.001 + 0.15 * float(i) / 4.0;
+    for (int i = 0; i < RAYMARCH_AO_SAMPLES; i++) {
+        float h = RAYMARCH_AO_MIN_DIST + RAYMARCH_AO_MAX_DIST * float(i) / (RAYMARCH_AO_SAMPLES-1);
```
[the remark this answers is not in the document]
patriciogonzalezvivo: Also note that `RAYMARCH_AO_SAMPLES ` is an `int` and in older GLSL versions there is not implicit casting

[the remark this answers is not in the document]
shadielhajj: @patriciogonzalezvivo it's not a myth. In this case, it's really a tiny micro-optimisation compared to the complexity of what we're doing, but why not.
Point taken re implicit casting of `int`
Pushed.

lighting/raymarch/softShadow.hlsl
```diff
@@ -30,22 +32,25 @@ options:
...
+float raymarchSoftShadow(float3 ro, float3 rd) {
+    const float mint = RAYMARCH_SHADOW_MIN_DIST;
+    const float maxt = RAYMARCH_SHADOW_MAX_DIST;
+    const float w = RAYMARCH_SHADOW_SOLID_ANGLE;
+
     float res = 1.0;
     float t = mint;
-    for (int i = 0; i < RAYMARCHSOFTSHADOW_ITERATIONS && t < maxt; i++)
-    {
+    for (int i = 0; i < RAYMARCHSOFTSHADOW_ITERATIONS; i++) {
```
patriciogonzalezvivo, line 42: Is there a need to do this in HLSL? It's only needed on `PLATFORM_WEBGL`

lighting/raymarch/softShadow.glsl
```diff
@@ -33,28 +32,25 @@ examples:
...
+    const float w = RAYMARCH_SHADOW_SOLID_ANGLE;
+
     float res = 1.0;
     float t = mint;
-    #if defined(PLATFORM_WEBGL)
     for (int i = 0; i < RAYMARCHSOFTSHADOW_ITERATIONS; i++) {
         if (t >= maxt)
             break;
-    #else
-    for (int i = 0; i < RAYMARCHSOFTSHADOW_ITERATIONS && t < maxt; i++) {
```
patriciogonzalezvivo, line 44: I feel that this is an optimization for non `PLATFORM_WEBGL` I know that makes is harder to read but probably more performant?

shadielhajj: I'm honestly not a fan of these micro-optimisations @patriciogonzalezvivo we might win a micro-second here (not even sure about that, we'd have to profile it). But we all lose several seconds everytime we have to read this.

I think this is part of a more general conversation about performance vs. readability. I see it this way:
* Lygia is not really an optimised library, it would take far more resources to make it so, systematic profiling, hardware specific optimisation, etc.. something completely outside of our reach for the moment. And IMO, these small optimisations will not make a significant difference.
* On the other hand it has a real educational vocation, as a reference library of shader knowledge. And these kind of constructs distract from this goal, by reducing readability.

Again, my personal opinion. But at the end of the day you're in charge of which direction the project takes and I'll follow your lead on this.

lighting/raymarch/softShadow.hlsl
shadielhajj: @patriciogonzalezvivo  You're right. But I'm trying to keep both versions as similar as possible. Makes it much easier for me to keep them in sync.

lighting/raymarch/softShadow.glsl
patriciogonzalezvivo: I will bring my [benchmarking tool back into shape](https://github.com/patriciogonzalezvivo/glslBenchmark). What if once the new feature are settle, the micro-optimizations are only for GLSL? I feel that having optimizations per-language is what make this very manual model worth it.

patriciogonzalezvivo: This is a great point. I will think on it

patriciogonzalezvivo (changes requested):

lighting/raymarch/softShadow.glsl
patriciogonzalezvivo: @shadielhajj I hear you. Per-language optimizations are a distraction and introduce friction. Specially for big re-factoring like you are doing right now. 
Also, is a reason to choose LYGIA over language translators. Is an ambitious goal, but hopefully one that keep us learning and improving. Keep us open to others to jump in to help and contribute. I prefer if we don't close the door to it. But also I don't want to slow you down. Let's not do micro optimizations at this stage. Once things settle down and we are not changing things all the time. When there is less need for agility. I can, if you are ok, came back to this and take on my self to apply them.

shadielhajj: Sounds good @patriciogonzalezvivo! Great to hear we're aligned on this. I would only add that when the time for optimisation comes, I recommend we do it using benchmarking/profiling. In my experience, the greatest bottlenecks are often not where one would expect them to be.

lighting/raymarch/softShadow.glsl
```diff
@@ -19,11 +18,11 @@ examples:
 #endif
 
 #ifndef RAYMARCH_SHADOW_MIN_DIST
-#define RAYMARCH_SHADOW_MIN_DIST 0.01
+#define RAYMARCH_SHADOW_MIN_DIST 0.005
 #endif
 
 #ifndef RAYMARCH_SHADOW_MAX_DIST
-#define RAYMARCH_SHADOW_MAX_DIST 3.0
+#define RAYMARCH_SHADOW_MAX_DIST RAYMARCH_MAX_DIST
```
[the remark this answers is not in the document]
shadielhajj: Done!

lighting/raymarch/softShadow.glsl
```diff
@@ -1,10 +1,9 @@
 #include "map.glsl"
-#include "../../math/saturate.glsl"
 
 /*
 contributors:  Inigo Quiles
 description: Calculate soft shadows http://iquilezles.org/www/articles/rmshadows/rmshadows.htm
-use: <float> raymarchSoftshadow( in <vec3> ro, in <vec3> rd, in <float> tmin, in <float> tmax) 
+use: <float> raymarchSoftshadow( in <vec3> ro, in <vec3> rd ) 
 options:
     - RAYMARCHSOFTSHADOW_ITERATIONS: shadow quality
```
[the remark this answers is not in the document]
shadielhajj: Done :-)

lighting/raymarch/softShadow.glsl
patriciogonzalezvivo: agree. Thinking on doing per function (file) test on different hardware.  Doing some first prototypes here: https://github.com/patriciogonzalezvivo/lygia_examples

shadielhajj:
@patriciogonzalezvivo Are we ready to merge this one?

PR #215, opened 2024-08-28
Cleaning small details for consistency

patriciogonzalezvivo:


lighting/material/add.glsl
```diff
@@ -9,7 +9,7 @@ license: MIT License (MIT) Copyright (c) 2024 Shadi EL Hajj
 #ifndef MATERIAL_ADD
 #define MATERIAL_ADD
 
-void add(Material a, Material b, Material r) {
+void materialAdd(Material a, Material b, Material r) {
```
patriciogonzalezvivo, line 12: I love function overloading. There is plenty on LYGIA. Because of WGSL lack of it, I think we should be more careful and try to be more unique on the naming of the functions. 
This fact will make the WGSL version very different, but we can reduce the warm

lighting/raymarch.glsl
```diff
@@ -42,8 +42,8 @@ license:
 #define RAYMARCH_CAMERA_FOV 60.0
 #endif
 
-#ifndef ENG_RAYMARCH
-#define ENG_RAYMARCH
```
patriciogonzalezvivo, line 46: I wasn't sure if this was a typo of there is a bigger plan behind this one

lighting/material/add.glsl
shadielhajj: Agreed!

lighting/raymarch.glsl
shadielhajj: I'm not sure I've done this 😅
There HUGE plans, but not on this specifically 😂
Feel free to rename.

shadielhajj (approved):

PR #217, opened 2024-08-28
Cook-Torrance BRDF

shadielhajj:
Started a full review of the PBR pipeline and stumbled on a strange mix-up:

Also took the opportunity to fix `pbrGlass` and `gooch` in HLSL, as well as a couple of missing includes.


patriciogonzalezvivo:
Hi @shadielhajj! The code looks good, doesn't break stuff and is nicer. But I'm getting lot of changes visually that I don't know if they are regressions or part of the transition. It is helpful if I added here?

shadielhajj:
@patriciogonzalezvivo Yes, the changes are expected. They should be improvements, in the sense of getting us closer to a correct PBR model, not regressions.
At some point I will do a systematic back-to-back with a reliable PBR engine, probably Filament or Unity, to confirm the correctness of the model. But I still have quite a few things to fix before we get there.
In the meanwhile, we have 2 options:
1. Merge the PRs progressively. And do the back-to-back when we're ready.
2. Hold off all PRs until the back-the-back is done, and accumulate all changes into a giant PR.
The disadvantage I see to the second approach, is that the PR will end up being massive, and we have to agree not to make major changes to the entire PBR system, in order to avoid merge conflicts.
Let me know what's your preference.

patriciogonzalezvivo:
I see. I need to focus on my 9-to-5 work. So if we go for option 2 I can promise that all PRs are holded by your by yours. 
And for those that are unrelated to the PBR system they will depend on you approval, ex: Metal or WGSL translations. 
For me option 2 will make it easier to spot/compare for actual regressions. So if you group the PRs for when visual quality or code review is useful the better. 

Do that sounds ok for you? if this any of it sounds like I pain. I can just keep making sure things compile on GLSL for native and WebGL.

shadielhajj:
Sounds like a plan @patriciogonzalezvivo! I'll put everything then in an going draft PR and we can review when it's ready!

patriciogonzalezvivo:
@shadielhajj I want to take the time to thank you not just for the code but also for the conversations and flexibility. We are in this process of knowing each other. I really want you to feel ownership and empower to bring your vision. So please keep bringing things up and speaking your mind freely. Thank you

shadielhajj:
Thanks for your trust and support @patriciogonzalezvivo! It's a massive project by any measure and we're all working on it in our free time. But I'm really excited about the possibilities and it's been so far a fun and insightful ride!

PR #219, opened 2024-08-28
PBR Overhaul

shadielhajj:
## Refactored Cook-Torrance

## Disney Roughness reparametrisation

We apply the classic Disney reparametrisation of roughness described in the Burley paper [Physically-Based Shading at Disney](https://disneyanimation.com/publications/physically-based-shading-at-disney/) including the clamping to 0.045 recommended by Sébastien Lagarde in [Moving Frostbite to PBR](https://www.ea.com/frostbite/news/moving-frostbite-to-pb). 
``` math
\alpha = perceptualRoughness^ 2
```
This allows for a better visual distribution of roughness throughout the entire range. The following comparison (from the Filament white paper) shows the new perceptually linear roughness (top) to the previous numerically linear roughness (bottom).
material_roughness_remap

## Reflectance Property

The Material struct currently includes an `f0` parameter which is very difficult to use (unless you're a light scientist). We replace it with an artist-friendly `reflectance` parameter. The parameter goes from 0 to 1 with 0.5 corresponding to an $$f0$$ of 0.04 (most dielectric material). The default value is fine for most cases but can be cranked all the way up to 1.0 if gems of high-gloss material are required.
Hi-Gloss

Combined with Metallic and Roughness, this adds a whole new palette of surface to play with.
From the Filament white paper again: From top to bottom: varying metallic, varying dielectric roughness, varying metallic roughness, varying reflectance.
material_parameters


shadielhajj:
@patriciogonzalezvivo this one depends on #217, so I'll wait until we sort that one out, before moving it to ready for review.

PR #222, opened 2024-09-22
Feature/metal blending

vade:
This PR adds two new sets of shaders

1 - Composite - Porter Duff composite functions and a top level header for the following operations:
   - compositeXor
   - destinationAtop
   - destinationIn
   - destinationOut
   - destinationOver
   - sourceAtop
   - sourceIn
   - sourceOut
   - sourceOver
   
 2 - Layer - this is Porter Duff Source Over compositing along with a Blend Mode. We implement RGBA 'layering' for all available RGB blend modes, extending them to RGBA with (what seems like) correct alpha semantics. We also include a main layer header.
 
 Note this only implements Composite and Layer for Metal for now. 
 
 This should have someone smarter than me™ review to ensure correctness!
 
 Also, we can leverage the Porter Duff composite operations in other shaders to make it semantically clear in the code what compositing we intend (a separate stream of work, just noting it)


patriciogonzalezvivo:
Wow @vade this is a fantstic PR! Will start reviewing it asap. Then try to do a GLSL translation.

Thank you!

math_compat.msl
```diff
@@ -1,6 +1,10 @@
 #ifndef METAL_COMPAT_FUNCS
 #define METAL_COMPAT_FUNCS
 
+inline float radians(float angle)
```
patriciogonzalezvivo, line 4: It's ok if we break this one into `/math/radians.msl` and the rest into `/math/lessThan.msl` and `/math/lessThanEqual.msl` ?

vade: Part of my thinking is theres a set of stupid missing features metal doesnt have that is out of the box in other languages? I dunno!

patriciogonzalezvivo:
Do you have a set of test code and images that you use for testing?

vade:
Nothing that I can share publicly right now tbh :(

vade:
Honestly it would be fairly trivial to port to GLSL if you wanted to test it there? most will just be renaming float3 / float4 to vec3 vec4 - there shouldn't be any other major issues I dont think :X

math_compat.msl
patriciogonzalezvivo: the `/math/` folder is ideal for that there is a lot of other polyfill functions in there, some important like this one other just for QOL like `frac/fract`, `lerp/mix`, `saturate`, etc. Other languages like for CUDA they implement all of them just because there is no native version

vade:
Im just a lazy fuck and I got what I wanted working ha :)

math_compat.msl
vade: Sure! we can do that. Do you think it makes sense to have a compat folder? My concern is it means that a ton of the existing ported metal shaders will need 1 by 1 different set of includes. 

I like the compat one shot include, but understand it may be different from the rest of the project. 

LMK.

vade:
Oh and BTW, the reason I named the blending + compositing modes Layer is it should be able to model Photoshop like layer blending methods (assuming you match the gamma / color space conventions). Just for the record.

patriciogonzalezvivo:
Lol I will merge it and do those changes and the translation then : ) 
thank you so much for all this pixel magic

Issue #225, opened 2024-10-05
intro to wgsl-tooling-wg 


mighdoll:
Hey, wanted to introduce our effort over in the [wgsl-tooling-wg](https://github.com/wgsl-tooling-wg/wesl-spec) to extend WGSL with community tooling. The idea is to extend WGSL to support features like `import` from local files and cargo/npm packages. We're also looking to add extensions for things like conditional compilation (`@if`), etc.  

It looks like Lygia could really use some standard tooling for `import`, and maybe conditional compilation.. Is there anything else Lygia would need?

patriciogonzalezvivo:
Hi @mighdoll! Thank you for reaching out! That's a great project! I will take a deeper look to it.

yes! some `import` feature would be great. I  went for using `#include` just because I saw several Rust parsers for WGSL that support `include` commands out of the box. It's a pretty old standard that is also adopted on all other shader languages and/or graphic engines. Of course if `import` becomes the standard I can change it.

Yes! some precompilation macros (defines and conditionals) would be amazing. As you can see in GLSL, HLSL and Metal, we use them extensively for templating, name collisions and back compatibility between different language version.  
We also use function defines also for versioning and templating. For example https://github.com/patriciogonzalezvivo/lygia/blob/main/sampler.glsl#L10

In my wish list for WGSL I also have function overloading! It allow us to build simple and friendly interfaces.

sdedovic:
+1 to both ideas, this looks awesome @mighdoll. Some of my thoughts:

A standard for imports / code splitting with conditional imports is much needed. An important request for WGSL extensions - they can not break vanilla code. I.e. if `import` or `@if(xxx)` statements are added to Lygia, I do not believe it should prevent the code from being used as-is with vanilla WebGPU APIs. This may mean either putting extensions inside of comments or having a new file extension, e.g. `wgslx` and keeping both vanilla WGPU and WGPU-ext code side-by-side. I think this is important as we should not introduce breaking changes that require third-party libraries.

--- 

While translating some files from GLSL to WGSL here, I ran into a few issues. Not all have solutions, but I'll enumerate anyway:

## 1. Function Overloading [gpuweb tracking issue](https://github.com/gpuweb/gpuweb/issues/876)
[...]

PR #233, opened 2024-11-02
Port some glsl shaders over to wgsl

josephreisinger:
This PR ports over a number of glsl shaders over to wgsl for usage in webgpu projects. 

As wgsl does not support function definition shadowing with different parameter type signatures, the naming convention roughly follows the practice of including a number after each function name to indicate the major parameter's dimension (e.g. func1 for `f32` vars and func3 for `vec3f`). Note that this may be controversial, e.g. in cases like `exposure3`.

Formatting is using `wgsl-analyzer`. 

A few important notes:
1. The lack of preprocessor #defines in wgsl means that we can no longer use define guards to ensure against circular dependencies or re-importing the same header. Combined with wgsl's lack of shadowing (noted above) I ended up having to patch `webpack-glsl-loader` (see fork here: https://github.com/josephreisinger/webpack-glsl-loader). This works for my particular case, but obviously is not helpful for folks who don't use webpack. Not sure what the best approach would be for this more generally / flagging for @patriciogonzalezvivo  
2. All these functions are have been tested *in situ* using the eyeball metric, but I haven't made any attempts to verify precision or otherwise test numerically.
3. Generally I have avoided making changes to existing files, except in a few small cases where necessary.

Please let me know if you have any feedback or if there is anything I can do to make this easier to review. Happy to split it up functionally or otherwise if that is better for you?


patriciogonzalezvivo:
Thank you @josephreisinger! great contribution! Wellcome permanently to the https://lygia.xyz/license as a contributor : ) 
I was wonder if you don't mind creating a `README_WGSL.md` with your translation notes + @sdedovic [notes](https://github.com/patriciogonzalezvivo/lygia/blob/main/DESIGN.md#wgsl-specifics). He also did some translations and arrive to similar solutions. There is also this [issue thread about WGSL constrains](https://github.com/patriciogonzalezvivo/lygia/issues/225) you will find interesting. Would be great if they can also hear you voice and wishes

josephreisinger:
Also I'm curious have you looked at [Naga](https://github.com/gfx-rs/wgpu/tree/trunk/naga) as a way to potentially automatically convert glsl->wgsl?

sdedovic:
Hey I'm a little late to the party. This looks awesome.

@josephreisinger In my initial attempts with Naga, I found it made the code quite unreadable and decided against it. I figure for something like this library, the ability to read and understand the code was important. I tried to keep variable names consistent between WGSL and the reference implementations in GLSL / HLSL. AFAIK, Naga renames variables. So I opted to do it mostly manually.

PR #236, opened 2024-12-07
Update cart2polar.glsl

jane00:
Add description for convention changed.

Add by jane00
Refer to https://en.wikipedia.org/wiki/Spherical_coordinate_system , there are two different definition for phi and theta, which are opposite to each other.

Before the changes by shadielhajj committed on 2024 Jul 26,  The physics convention(ISO convention) is followed, phi means "azimuthal angle", theta for "polar angle"
but now, The "mathematics convention" is followed, phi means "polar angle", theta for "azimuthal angle"

So be careful with the old code !!!

### DIFF
--- a/space/cart2polar.glsl
+++ b/space/cart2polar.glsl
@@ -11,6 +11,16 @@ vec2 cart2polar(in vec2 st) {
     return vec2(atan(st.y, st.x), length(st));
 }
 
+/*
+Add by jane00
+Refer to https://en.wikipedia.org/wiki/Spherical_coordinate_system , there are two different definition for phi and theta, which are opposite to each other.
+
+Before the changes by shadielhajj committed on 2024 Jul 26,  The physics convention(ISO convention) is followed, phi means "azimuthal angle", theta for "polar angle"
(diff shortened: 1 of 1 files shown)


shadielhajj (approved):

patriciogonzalezvivo:
@jane00 feels like this description belongs to the 'description' section on the YAML formatted header (in comments). And should be added to: 'cart2polar.glsl', 'cart2polar.hlsl', 'cart2polar.msl' and also 'polar2cart.glsl', 'polar2cart.hlsl' and 'polar2cart.msl'

jane00:
> @jane00 feels like this description belongs to the 'description' section on the YAML formatted header (in comments). And should be added to: 'cart2polar.glsl', 'cart2polar.hlsl', 'cart2polar.msl' and also 'polar2cart.glsl', 'polar2cart.hlsl' and 'polar2cart.msl'

Sorry, I'm not very familiar with the correct format conventions, so my submission might not be very standardized. However, I happened to use this function. I originally wanted to add a "vec3 polar2cart(in vec3 polar)" function myself, but I found that the new version already includes it. Then I noticed the definition had changed, so I added a note about it.

PR #250, opened 2025-02-26
Make MSL mod() implementation match GLSL

haysmike:
# Problem

`mod()` in Metal shaders is 0 for negative inputs because it's using `fmod()`.

# Solution

Implement using `floor()`, the same way it's implemented for HLSL.

# Testing

I created a simple fragment shader with the following code:

```c++
    float3 modColor = float3(mod(st.x, 1));
    float3 fmodColor = float3(fmod(st.x, 1));
    float3 color = mix(fmodColor, modColor, step(0, st.y));
    return float4(color, 1);
```

When centered at 0,0 and y in (-1,1), this generates:

image

Similar test for `float2` version:

image

### DIFF
--- a/math/mod.msl
+++ b/math/mod.msl
@@ -1,7 +1,7 @@
 /*
-contributors: Patricio Gonzalez Vivo, Anton Marini
-description: mod polyfill
-use: <bool2|bool3|bool4> mod(<float2|float3|float4>, <float2|float3|float4>) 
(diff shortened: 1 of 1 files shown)


vade:
Oh nice Catch!

vade (approved):
Great stuff! Thank you!

@patriciogonzalezvivo i'll approve this, but wont merge - not sure the best flow here.

haysmike:
Thanks for the quick review @vade!

PR #258, opened 2025-05-04
package lygia/math with wesl packaging tool

mighdoll:
This patch publishes lygia/math functions to npm in a form so that users can easily import lygia using the wesl tools. With this, users can e.g. `import lygia::math::rotate3d;` in their shader code. 

wesl supports a variety of [build configurations](https://wesl-lang.dev/docs/JavaScript-Builds) including vite and webpack.  And forthcoming wgsl/wesl tools like syntax highlighters and language servers will support this syntax and packaging too. 

I wrote a quick example here https://github.com/mighdoll/lygia-example of using lygia in a shader.
(lygia-example should be in the same directory as lygia - it uses a link: dependency until lygia is published.)

Changes:
- New bits in lygia's package.json to build and expose the wesl bundles.
- The changes in the shaders from .wgsl to .wesl are mostly just for the import syntax. i.e. mechanically changing `#include "../math/mod289.wgsl"` to `include lygia::math::mod289;`. WESL just extends WGSL, so any future WGSL shader code should just work.
- a couple of wgsl syntax / formatting tweaks
- renamed `const.wesl` to `consts.wesl` to avoid name conflict with the `const` keyword in WGSL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "rm -rf dist/* && webpack",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "build:wesl": "wesl-packager --src \"math/*.wesl\" --projectDir . --rootDir . --updatePackageJson --multiBundle --outDir dist",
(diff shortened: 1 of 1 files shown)


patriciogonzalezvivo:
Fantastic work @mighdoll ! I'm excited to open this new chapter!

PR #261, opened 2025-05-15
Added Animation & Easing WGSL Shaders

UstymUkhman:
WGSL animation shaders were adapted from their GLSL counterpart.
Code was tested only for compilation at the folowing [online resource](https://takahirox.github.io/online-wgsl-editor/index.html).


patriciogonzalezvivo:
Thank you @UstymUkhman!

PR #270, opened 2025-09-20
WESL fixes

k2d222:
* changed `import lygia::` to `import package::`
* fixed some missing/incorrect imports
* some code fixes (I'm sure there is more to fix, this is all the tool could detect)
* formatted files


math/unpack.wesl
```diff
@@ -44,7 +44,7 @@ fn unpack256(v: vec3) -> f32  {
     return dot(v, f) / 16581375.0;
 }
 
-fn unpack(v: vec3f, base: f32) -> f32 {
+fn unpackBase(v: vec3f, base: f32) -> f32 {
```
k2d222, line 47: I hope WESL will support soon overloads so we don't have to deal with this anymore

color/space/ryb2rgb.wesl
```diff
@@ -15,23 +15,23 @@ fn ryb2rgb(ryb: vec3f) -> vec3f {
...
   let ryb101 = vec3f(.5, 0., .5);       // Violet
   let ryb110 = vec3f(1., .5, 0.);       // Orange
   let ryb111 = vec3f(0., 0., 0.);       // Black
   return 
     cubicMix3(
       cubicMix3(
-        cubicMix3(ryb000, ryb001, ryb.z),
-        cubicMix3(ryb010, ryb011, ryb.z),
-        ryb.y
+        cubicMix3(ryb000, ryb001, vec3f(ryb.z)),
```
k2d222, line 26: not ideal because the WESL files don't have the `cubicMix(vec, vec, f32)` overloads.

math/unpack.wesl
patriciogonzalezvivo: I'm so looking forward to it

PR #276, opened 2025-11-03
finish WESL color port and tests

mighdoll:
- note that rgb2xyz.wesl scales by 100 unlike the glsl #276
- I've added some comments that gamma, DITHER_BLUENOISE_TIME
  and similar constants should be made user configurable later,
  once WESL implements '@param const' (planned for next release).
- shows off a new kind of test: image snapshot tests, e.g. for dither

### DIFF
--- a/test/wesl/__image_snapshots__/dither-bayer3-gradient.png
+++ b/test/wesl/__image_snapshots__/dither-bayer3-gradient.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:918f475e9973043175ecb34263f3e030000e3de6826f6360b155f0e711d142d8
+size 19654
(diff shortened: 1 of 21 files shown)


mighdoll:
converted PR to draft mode, while we figure out a new approach to reference images

patriciogonzalezvivo:
Hi @mighdoll ! let's go the git-lfs route. If I understand well, that will not impact (with longer download time and more space) non-dev users. Right?

Thank you so much @mighdoll for researching this option! 
lygia.xyz have already some scripts for uploading files. I will need to add some security on it. But is good I can do that slowly and not loose momentum.

@vade we talk with @mighdoll couple months back on adding a test folder into the main mono-repo. We think this is the most practical way forward. Please feel free to make the Metal test yours. 

For GLSL I have been using also this image comparison approach using this repo  https://github.com/patriciogonzalezvivo/lygia_examples please use all the examples you need. Let's plan for some convergence on testing. Getting exactly the same images could be tricky, but worth the shot.

mighdoll:
closing this, merged the squashed version #277.

PR #278, opened 2025-11-05
WESL port of space/ with tests

mighdoll:
uses git-lfs too

this PR includes two commits. The first is the color/ PR (because of how github PRs work, grr.) I'm hoping the first commit will magically disapper when the prior one is merged..

In the meantime if you're in a reviewing mood, just look at the second space/ commit here.


mighdoll:
The first commit doesn't magically disappear from this PR even after it was merged, so no magic solution there to stacking PRs. Oh well, maybe someday github will fix things. 

So if reviewing this PR in the future, note that only the second commit is new stuff (the first commit is an earlier PR).

Issue #288, opened 2025-11-10
`psrdnoise()` suffers from truncation effects


haysmike:
I'm noticing what I think are referred to in the code as "truncation effects" in `psrdnoise()`:

<img width="888" height="501" alt="Image" src="https://github.com/user-attachments/assets/b0f64594-7d97-4214-92af-5dda5b2e201d" />

Code is here for reference:

```C
#include "../lygia/generative/psrdnoise.glsl"

void mainImage(out vec4 fragColor, in vec2 fragCoord) {
  vec2 uv = fragCoord/iResolution.x;
  vec3 color;
  psrdnoise(vec3(uv * 16.0, iTime), vec3(0), 1.0, color);
  fragColor = vec4(color, 1.0);
}
```

I've done a bit of debugging, and found that the `permute()` function appears to be the culprit. In stegu's [reference implementation](https://github.com/stegu/psrdnoise/blob/419175a270862ce7ae692038fafafb42ec0427e9/src/psrdnoise3.glsl#L65-L68), the function looks like:

```C
vec4 permute(vec4 x) {
     vec4 xm = mod(x, 289.0);
     return mod(((xm*34.0)+10.0)*xm, 289.0);
}
```

Lygia's is a bit simpler:
```C
vec4 permute(const in vec4 v) { return mod289(((v * 34.0) + 1.0) * v); }
```

When I replace the `permute()` function with stegu's, I get the correct behavior, but:
- I don't know if we want to change `permute()` for all Lygia use-cases
- ~I have no idea why the same implementation works fine in Metal~ Metal has the same behavior
- There are 6 calls to `permute()` in the `psrdnoise()` implementation, so it would be nice to extract something (although simply replacing with `permute(mod289())` does work)

~I'd love to put up a PR but could use some guidance here! Let me know what you think about the best way to approach this.~ PR is up! Thanks!!

PR #290, opened 2025-11-11
Fix truncation issues in psrdnoise()

haysmike:
Fixes https://github.com/patriciogonzalezvivo/lygia/issues/288.

I tested the Metal and GLSL implementations with large values. I don't have an easy way to test HLSL. There is no WGSL implementation yet.


patriciogonzalezvivo:
Thank you @haysmike!

PR #314, opened 2026-09-07
Make graident noise configurable

nwhetsell:
This PR attempts to add macros to configure the noise functions used for gradient noise. I tried to base this off of the FBM files (for example, [generative/fbm.glsl](https://github.com/patriciogonzalezvivo/lygia/blob/370bb5081ddca28a9c6199cf4673e365f61224fb/generative/fbm.glsl)), but please note that I’m able to test only GLSL files.


patriciogonzalezvivo:
This is great @nwhetsell ! Thank you 🙏

